PR #214, opened 2015-02-25
Add docs for the Rule class

sheyll:
More documentation for the "Rule" type class; When users coming from
"make" approach shake they might have some misconceptions about 
the role of "Rule" instances, especially about the relation of "key" and
"value". It is obvious to me now, but I had a hard time understanding 
the actual role of that class, and how to create reasonable custom 
instances.
This is only a draft - are documentation pull requests at all desired?
What is the preferred style of documentation?
Please comment on the request.

### DIFF
--- a/src/Development/Shake/Core.hs
+++ b/src/Development/Shake/Core.hs
@@ -122,11 +122,10 @@ type ShakeValue a = (Show a, Typeable a, Eq a, Hashable a, Binary a, NFData a)
 --  It is important to distinguish 'Rule' instances from actual /rules/. 'Rule'
 --  instances are however necessary, for the creation rules.
 --
---  To add build rules use the 'rule' function to specify the 'Action' that
---  builds an artifacts identified by a @key@.
+--  Actual /rules/ are functions from a @key@ to an 'Action'; they are
+--  culminated into a 'Rules' monad using the 'rule' function.
 --
---  Refering to the example above, this is how an actual /rule/ could be
---  created:
+--  Refering to the example above, this is how a rule can be created:
 --
 --  @
 --      -- Compile @foo@ files; for every @foo@ output file there must be a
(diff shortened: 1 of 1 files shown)


ndmitchell:
Pull requests are most appreciated, thanks for takin the time, and for pointing out what is confusing at the beginning! I'm just heading to bed now, but will read through the details tomorrow morning. 

Generally I don't expect most people to create their own rules, but just use the baked in ones. What did you need to create a custom rule for?

sheyll:
I actually don't ever compile any source code with shake. I use shake to orchestrate individual builds running on vms.

I plan on extending the use of shake to actual compilation of haskell and C code currently build using cmake; I only wish there was a library of helpers for compiling and linking C projects into say rpm files.

The ultimate-result of my shake build is a set virtual machine images and a bunch of packages grouped in  (immutable) package repositories. The basic idea is that several (related) packages are built are grouped into package repositories.
Whenever a single package of a group changes, a new repo is created (i.e. the package files are copied to the package repository directory in a new subdirectory and new yum repo files etc are created)

I have three custom rules, not all of which might necessarily be implemented as such:
 1. VM image builds
 2. Package repository builds
 3. Cloning/Pulling of git repositories.

I am sure I want a custom rule to build or pull VM images using my experimental "b9" library (at http://hackage.haskell.org/package/b9-0.2.4), because b9 has the concept of "shared images" which can be reused by other vm-builds but also can be "pushed" or "pulled" to/from remote hosts.

These images are hidden in a local directory known to b9  - and are never addressed by any build b9 build directly, e.g. through a filesystem path, but only via an abstract identifier, e.g. "rh7-prod"  or "fc20-testing". 
One reason why I cannot allow a build to directly specify a file system path to an image file(-set) is that a build job might actually update an such an image while another build job on the same machine copies that images.

ndmitchell:
Thanks for the changes. I hadn't expected many people to be writing their own Rules, but your uses seem quite reasonable - very interesting to know.

I have some fairly extensive documentation tests which check all the Haddock samples compile. I'm just fixing the docs up so they pass those (both fixing the docs, filling out types for your examples, and fixing the doc generator). I should have that done tonight, then tomorrow I'll take a closer look at the actual content of the docs.

ndmitchell:
OK, all the docs pass, and I've gone through and revised the Rule instance. It's pretty much following the pattern you started with, but I've tweaked the example to just be defining the "file" rules, and reworded things a little. Let me know if you think I've made it worse in any way. Thanks for taking the time to improve the docs here, I think it's quite valuable. I even found a minor bug in one of the comments on `storedValue`, and inserted a test case to confirm how it actually works, so a very valuable exercise.

sheyll:
Wow, I am impressed by the sophistication of your approach. Thanks for
documenting shake a little more.

Wbr
Sven
Am 28.02.2015 23:23 schrieb "Neil Mitchell" <notifications@github.com>:

PR #215, opened 2015-02-26
Fix documentation

sheyll:
Most important change is to add a missing "NOT"


ndmitchell:
Thanks, I merged it with the changes I was making at the time, now pushed.

PR #233, opened 2015-04-17
Fix typo in Manual.md

enolan:
Three times, the manual says to run `runhaskell _build/run` to run the build system. That's the final executable to be built in the example.  `runhaskell Build.hs` is the Haskell file with the build system in it.


ndmitchell:
Thanks for spotting that! I've also reviewed it and found I didn't mention `runhaskell Build.hs` anywhere, so I added that near the top. I'll update the website later today.

ndmitchell:
All updated on the website now.

PR #245, opened 2015-05-12
Adding support for fsatrace

jacereda:


ndmitchell:
All looks thoroughly reasonable. The only issue I spotted is that it doesn't compile on older GHC's as they don't have `setEnv`. Given that Shake is an API as well, using `setEnv` is usually a bad plan, since it is a global setting and isn't undone afterwards. Fortunately, I think the setting of environment variables can be pushed down so it only happens when spawning a process. I've merged your branch locally, will do some stuff to it tomorrow, and then aim to push it by midday.

ndmitchell:
A bit more work than I expected, but I've now merged it. Instead of setEnv I added `AddEnv` directly to the command running stuff, and then used it here. Can you please check I didn't break anything, and that it all seems to match what you were expecting?

jacereda:
I switched from Haskell Platform to GHC for Mac OS X (7.10.1) over the weekend and I'm now building from a sandbox. It seems the test suite has problems when running sandboxed. In src/Test/Self.hs I had to do:

            sandboxed <- doesFileExist "cabal.sandbox.config"
            if sandboxed
            then cmd "cabal" "exec" "--" "ghc" flags args
            else cmd "ghc" flags args

But then I get more errors (runhaskell-related I think). It could well be my ignorance regarding haskell tools, am I missing something? Should I just give up the sandbox thing?

ndmitchell:
The easiest thing is to avoid sand boxing. I know some people have successfully tested using the sandbox, but I don't know what special steps are required.

jacereda:
Ok, got it running, opened a separate PR for the problems I got regarding ./foo/bar vs foo/bar.

PR #246, opened 2015-05-19
Normalise paths reported by fsatrace, whitelist ~/.cabal and ~/Applications

jacereda:
With this patch the test suite passes when running under fsatrace.

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -203,8 +203,8 @@ fsatraceFiles file = do
     xs <- parseFSAT <$> readFileUTF8 file
     let reader (FSATRead x) = Just x; reader _ = Nothing
         writer (FSATWrite x) = Just x; writer (FSATMove x y) = Just x; writer _ = Nothing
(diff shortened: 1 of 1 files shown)


ndmitchell:
Awesome!

ndmitchell:
Does fsatrace work under Linux yet? The next step is to get it working there, and turned on in the buildbot, so we can keep the test suite running under tracker.

jacereda:
I can add the #ifdefs that I had in an older version that ran on Linux at some point if you have access to a linux box to test. Otherwise I can try to find some time at work (not this week probably) and try to test it there.

ndmitchell:
I have a Linux VM I can try on, but I tend to just hope and submit to the Travis Linux CI service and debug it that way.

jacereda:
Just pushed https://github.com/jacereda/fsatrace/commit/4dde001c56321359d145ce0d2ff3167c28798295

On linux you could try it as:

    env FSA_OUT=/dev/stdout LD_PRELOAD=fsatrace.so cp /bin/ls /tmp/foo

jacereda:
Sorry, that should read FSAT_OUT.

Issue #270, opened 2015-06-25
newCache can't query itself


nh2:
I noticed that I can't write a `Rule` like this using `newCache`:

```haskell
{-# LANGUAGE ScopedTypeVariables #-}
... = do -- :: Rules

  (getRecursiveIncludesOf :: FilePath -> [FilePath]) <- newCache $ \file -> do
    includes <- scanIncludesSingleFile file
    allIncludes <- traverse getRecursiveIncludesOf includes
    return allIncludes
```

because I'm in the body of the rule that defines `getRecursiveIncludesOf` (the name is not in scope inside the body).

However, I think for `newCache` this is sensible to have.

I wonder whether we should have a version of `newCache` that passes the cached function in.

--

Alternatively, I wonder if `Rules` should have a [`MonadFix`](http://hackage.haskell.org/package/base-4.8.0.0/docs/Control-Monad-Fix.html#t:MonadFix) instance! Then we could use `recursive do` notation (or `mdo`, see section *7.3.11. The recursive do-notation* in [the docs](https://downloads.haskell.org/~ghc/7.8.4/docs/html/users_guide/syntax-extns.html#7.3.8.%20The%20recursive%20do-notation)):

```haskell
{-# LANGUAGE RecursiveDo #-}
{-# LANGUAGE ScopedTypeVariables #-}
... = do -- :: Rules

  rec {
    getRecursiveIncludesOf <- newCache $ \file -> do
      includes <- scanIncludesSingleFile file
      allIncludes <- traverse getRecursiveIncludesOf includes
      return allIncludes
  }
```

It seems to me that it makes sense for `Rules` to have a `MonadFix` instance, since rules never run anything directly - they build a plan of things that can be built, and things inside (the `Action`s) are fundamentally allowed to ask for files built by the rules that implement the rule. The way around this (depending on a rule from inside the same rule) is usually: Because we use `Strings` in `need`!

The fact that `Rules` are order-independent supports this idea, and makes me think that it'd make sense for rules to be written in an `mdo` block in general (but that's a user's choice).

nh2:
In fact, we can just `newtype Rules a = Rules (WriterT (SRules Action) IO a) deriving (..., MonadFix)`!

I added a PR, but let's first discuss whether this makes sense.

ndmitchell:
[...]

PR #271, opened 2015-06-25
Add `instance MonadFix Rules`. Can be auto-derived

nh2:
Tentative. Discussion in https://github.com/ndmitchell/shake/issues/270


ndmitchell:
I thought for a long time, and ultimately decided that basically ContT is the only thing that doesn't support MonadFix (and Free monads, but there the implementation on top usually can, and List monads, but they have lots of weird bits anyway). Since Rules are really a commutative/associative bag of things, having MonadFix makes some sense, and it is derivable and likely to remain so for all future iterations of Rules, so it isn't going to hurt too much.

PR #274, opened 2015-07-02
add doc link for FilePath

ggreif:
Not sure whether there is a better link target for this. Hackage (latest) is good enough for me.


ndmitchell:
Thanks! Good idea.

Issue #202, opened 2014-12-04
Config module is weak


ndmitchell:
The Config module doesn't support generated the config at runtime, and the Ninja lexical syntax means that all include statements are relative to the current directory, not the file doing the including, so you can't relocate directories.

I think if `usingConfig` took a `Action (Map String String)` as the first argument then everything would generalise nicely. You could then have better `readConfig` functions, but they would be entirely optional.

ndmitchell:
See #279 which does a bit of work towards this.

PR #279, opened 2015-07-18
Add support for generated configuration files

snowleopard:
A simple fix to support generated configuration files in usingConfigFile. Already mentioned in issue #202.


ndmitchell:
Thanks. A useful first step to doing something sensible with generated config files.

PR #303, opened 2015-09-03
Rule names may contain dots.

awson:
After very recent changes in CVS LLVM/Clang, cmake generated ninja project (perhaps make variant is vulnerable too) can't be built by shake. The patch proposed fixes things for me.


ndmitchell:
Thanks! I extended my test suite with an example, and confirmed that I match the real Ninja behavior - in particular that `foo.bar` is a valid variable name, but `$foo.bar` is equivalent to `${foo}.bar`.

PR #306, opened 2015-09-25
Support newer fsatrace and sandboxed suite

jacereda:


jacereda:
@ndmitchell  do you think the failed build is related to these changes?

ndmitchell:
@jacereda - no, GHC pre 7.8 has some race conditions in their file-handle table, and every so often the Shake test suite trips over it. Nothing I can do but hit the retry button and it usually goes away. Just reviewing the patches now.

src/Development/Shake/Command.hs
```diff
@@ -219,19 +220,31 @@ parseFSAT = mapMaybe (f . wordsBy (== ':')) . lines
           f _ = Nothing
 
 
-unixWhitelist :: IO [FilePath]
-unixWhitelist = do
+unixWhitelist :: [FilePath]
+unixWhitelist = unsafePerformIO $ do
     home <- getEnv "HOME"
```
ndmitchell: Should this be [`getHomeDirectory`](https://hackage.haskell.org/package/directory-1.2.4.0/docs/System-Directory.html#v:getHomeDirectory)?

src/Development/Shake/Command.hs
```diff
@@ -140,34 +140,35 @@ commandExplicit funcName copts results exe args = do
...
-                res <- act vars exe args
+                res <- act fsat $ file:"--":exe:args
                 (rs, ws) <- liftIO $ fsatraceFiles file
-                whitelist <- liftIO unixWhitelist
-                let whitelisted x = any (\w -> (w ++ "/") `isPrefixOf` x) whitelist
-                trackRead $ filter (not . whitelisted) rs
-                trackWrite $ filter (not . whitelisted) ws
+                let whitelisted x = any (`isPrefixOf` x) unixWhitelist
+                    ham = filter (not . whitelisted)
+                    rel = mapM (liftIO . makeRelativeToCurrentDirectory)
```
ndmitchell: `makeRelativeToCurrentDirectory` is literally defined as `makeRelative` on `getCurrentDirectory`, so I would suggest pulling out `getCurrentDirectory` once, and then calling `makeRelative` many times, which makes this step pure and likely faster.

src/Development/Shake/Command.hs
```diff
@@ -219,19 +220,31 @@ parseFSAT = mapMaybe (f . wordsBy (== ':')) . lines
...
-           ,"/var/folders"
-           ,"/usr"
-           ,"/Library"
-           ,"/System"
-           ]
+    ghcPath <- lookupEnv "GHC_PACKAGE_PATH"
+    sbcPath <- lookupEnv "CABAL_SANDBOX_CONFIG"
+    cabPath <- lookupEnv "CABAL_SANDBOX_PACKAGE_PATH"
+    let sep = if isWindows then ';' else ':'
+        splitted = map dropFileName . wordsBy (== sep) . fromMaybe ""
```
ndmitchell: It's better to use `splitSearchPath` from the filepath library which does the splitting and always does the right thing on Windows vs Linux.

src/Development/Shake/Command.hs
```diff
@@ -219,19 +220,31 @@ parseFSAT = mapMaybe (f . wordsBy (== ':')) . lines
...
+                    ,"/Library/"
+                    ,"/System/"
+                    ,"/private/var/"
+                    ,"/etc/"
+                    ,"/opt/"
+                    ,"/lib/"
+                    ]
+        sandbox = case sbcPath of
+          Nothing -> []
+          Just x -> [x]
```
ndmitchell: This is just an instance of `sandbox = maybeToList sbcPath`.

src/Development/Shake/Core.hs
```diff
@@ -541,8 +541,10 @@ applyKeyValue ks = do
                 liftIO $ globalLint $ "before building " ++ top
                 putWhen Chatty $ "# " ++ show k
                 res <- runExecute globalRules k
-                when (shakeLint globalOptions == Just LintTracker)
-                    trackCheckUsed
+                case shakeLint globalOptions of
+                  Just LintTracker -> trackCheckUsed
+                  Just LintFSATrace -> trackCheckUsed
+                  _ -> return ()
```
ndmitchell: All other indentation in Shake is 4 spaces, so best to be consistent.

src/Test/Command.hs
```diff
@@ -51,7 +51,8 @@ main = shaken test $ \args obj -> do
             ]
 
     obj "shake_helper.hs" %> \out -> do need ["src/Test/Command.hs"]; writeFileChanged out helper_source
-    obj "shake_helper" <.> exe %> \out -> do need [obj "shake_helper.hs"]; cmd (Cwd $ obj "") "ghc --make" "shake_helper.hs -o shake_helper"
+    [obj "shake_helper", obj "shake_helper.o", obj "shake_helper.hi"] &%> \_ -> do
+      need [obj "shake_helper.hs"]; cmd (Cwd $ obj "") "ghc --make" "shake_helper.hs -o shake_helper"
```
ndmitchell: Not sure why this needed changing? It seems unrelated, and at first glance looks like it no longer works on Windows (it used to build `shake_helper.exe`, now it just builds `shake_helper`).

src/Test/Docs.hs
```diff
@@ -21,15 +21,14 @@ main = shaken noTest $ \args obj -> do
 
     config %> \_ -> do
         need ["shake.cabal"]
-        unit $ cmd "runhaskell Setup.hs configure" ["--builddir=" ++ obj "dist","--user"]
+        unit $ cmd "cabal configure" ["--builddir=" ++ obj "dist","--user"]
         trackAllow [obj "dist//*"]
 
     index %> \_ -> do
-        need [config,"shake.cabal"]
+        need [config,"shake.cabal","README.md","CHANGES.txt"]
```
ndmitchell: Why the extra needs? I don't see a use for them.

src/Test/Docs.hs
```diff
@@ -21,15 +21,14 @@ main = shaken noTest $ \args obj -> do
 
     config %> \_ -> do
         need ["shake.cabal"]
-        unit $ cmd "runhaskell Setup.hs configure" ["--builddir=" ++ obj "dist","--user"]
+        unit $ cmd "cabal configure" ["--builddir=" ++ obj "dist","--user"]
```
ndmitchell: Is moving to `cabal` required for running in a sandbox? I think I used to use `cabal` instead of `runhaskell`, but then it started not working and I switched. Generally `cabal` would be preferable if it always worked.

src/Test/Lint.hs
```diff
@@ -94,11 +94,11 @@ main = shaken test $ \args obj -> do
         need [obj "tracker-source.c", obj "tracker-source.h"]
         cmd "gcc" ["-c", obj "tracker-source.c", "-o", out]
     where gen t f = unit $ if isWindows
-                           then cmd "cmd /c" ["echo" ++ t ++ " > " ++ f]
-                           else cmd Shell "echo" ["x", ">", f]
+                           then cmd "cmd /c" ["echo " ++ t ++ " >" ++ f]
+                           else cmd "sh -c" ["echo " ++ t ++ " >" ++ f]
```
ndmitchell: Doesn't `Shell` do roughly `sh -c`? Although I guess they should either both use `Shell` for the test, or neither use it (and here, I think actually they're sufficiently similar that `cmd Shell ["echo " ++ t ++ " > " ++ f]` might cover both).

src/Test/Lint.hs
```diff
@@ -117,8 +117,7 @@ test build obj = do
     crash ["--clean","listing","existance"] ["changed since being depended upon"]
     crash ["needed1"] ["'needed' file required rebuilding"]
     build ["needed2"]
-
-    when False $ whenM hasTracker $ do
+    whenM hasTracker $ do
```
ndmitchell: Awesome! I've had a TODO: get this working in my local repo for ages, so a great result.

ndmitchell:
Sorry for taking a while to get round to reviewing this, it took a while, and I didn't have much time over the weekend until now. A very good patch, very happy with it, and very grateful that you got the test suite fixed it - something that's been on my todo list a while. Of the comments, src/Test/Command.hs is probably important to fix before merge or immediately after, and the other are all little tweaks to refactor after. I expect to merge it tomorrow sometime (if you reply before then or have updates or more info, great, otherwise I'll merge it and do the tweaks myself).

The test suite eventually passed - it did seem worse than normal that time, so I'll have to see why.

jacereda:
Ok, I applied most of your suggestions and will comment on the rest after the suite runs.

src/Test/Command.hs
jacereda: Looks like I messed the '<.> exe' part. Shouldn't the intermediates be listed as products? Otherwise the linter will complain...

src/Test/Docs.hs
jacereda: Isn't haddock copying those to the build directory? I added those because the linter complained.

src/Test/Docs.hs
jacereda: I run the suite inside the sandbox as:

    env FSAT=<path-to-fsatrace> cabal exec -- cabal test

I can try again, but I think runhaskell isn't affected by CABAL_SANDBOX_PACKAGE_PATH.

src/Test/Lint.hs
jacereda: This one I wanted to discuss with you. Shell isn't right now working properly because the code injection happens in the fsatrace function. That is, Shell will execute the "fsatrace <output> -- whatever" command, but we need the "sh" part after "fsatrace". 

In any case, shouldn't the shell to use be left to the user? I mean, a user could want to have some parts of the build system use "cmd" and some others use "bash" or "zsh" or whatever. Maybe Shell could just disappear.

src/Test/Docs.hs
ndmitchell: Sure, sounds reasonable - they aren't required, but best make the linter happy.

src/Test/Docs.hs
ndmitchell: OK, makes sense, I'm happy with that. I think there were some screwy cabal installs going around at one point that pushed me to runhaskell, but people can always upgrade cabal.

src/Test/Lint.hs
ndmitchell: I don't want Shell to disappear, it's quite handy, since it's a built in part of the Haskell process library, and can be quite useful, and avoids an isWindows. But reworking it so I do either `sh -c` or `cmd /c` in a portable way seems valuable - so instead of passing `Shell` to Haskell to deal with, I do it myself, then `fsatrace` should work fine. I've ticketed it here: https://github.com/ndmitchell/shake/issues/308

ndmitchell:
OK, I'm happy to merge with that, unless you have any more comments you wish to make given your successful test run?

jacereda:
Great. I don’t like the whitelisting part, but I’ll probably open a separate PR for that.

Are you ok with letting the user choose whatever approach seems handy in his concrete use case (either whitelisting or blacklisting)? In my case, I would just ignore anything that doesn’t fall inside the sources or build directories and that can be tested quite fast.

ndmitchell:
I think I'd prefer blacklisting - generally you only care about dependencies that are in your source tree (a single specific directory), and then anything else you want to add tends to be thought about separately. As an example, you typically want to depend on `gcc --version` rather than the timestamp of `gcc.exe` and all libraries it reads in.

However, if you want both, I'm happy to support both.

PR #315, opened 2015-10-09
Fix link in index page

kirbyfan64:
I was so confused why I kept ending up at samplecount.com... :)


ndmitchell:
Thanks! I will regen the page over the weekend.

ndmitchell:
Regenerated now - thanks!

PR #319, opened 2015-10-16
Get rid of the hardcoded lint whitelist

jacereda:


src/Development/Shake/Command.hs
```diff
@@ -143,14 +143,19 @@ commandExplicit funcName copts results exe args = do
...
+          trackWrite $ rel $ ham ws
 
         winTracker act = do
             (dir, cleanup) <- liftIO newTempDir
             flip actionFinally cleanup $ do
                 res <- act "tracker" $ "/if":dir:"/c":exe:args
-                (rs, ws) <- liftIO $ trackerFiles dir
-                trackRead rs
-                trackWrite ws
+                liftIO (trackerFiles dir) >>= track
```
jacereda: I'm not sure about this... Does tracker.exe report absolute paths? If not, this should probably be piped through `makeAbsolute` prior to calling `track`.

ndmitchell:
Tracker spits out absolute paths. However, with your tool, I'm less fussed about tracker and would like to remove it at some point.

Two thoughts about `shakeLintFilter`:

* Should it be a list of `FilePath`? Or, perhaps, `FilePattern`? There is `trackAllow` which serves as the inverse, so people can focus to the current directory with `shakeLintFilter`, then allow specific extensions etc with `trackAllow`.
* The name doesn't really imply the "sense" - do we look at filtered things, or ignore them? `shakeLintFocus` is the best I can come up with so far, but that's also not very good. Any thoughts?

jacereda:
`shakeLinted`? `shakeLintOnly`?
As for `FilePath`/`FilePattern`, I guess the second would make more sense since that's what `trackAllow` uses.

ndmitchell:
shakeLintInside and use FilePath might work. Not sure FilePattern is as necessary - provided you can restrict to a directory the allow give you patterns and thus enough power to refine it. Unless you can think of examples where we want more? I'm also aware that going from Path to Pattern is easy, but doing the reverse is a breaking change.

jacereda:
Sounds good.

jacereda:
On a second thought, I think a negative form might also be useful. Something like `shakeLintIgnore`. The use case is present in the current `shakeLintFilter` usage, we want all things hanging from the current directory tracked but things like .cabal-sandbox/ should be ignored. If there're lots of rules using cabal it would be cumbersome to add explicit `trackAllow`s in each rule. What do you think?

jacereda:
I implemented both options, not sure the option names are good though...

ndmitchell:
Thinking further, I agree, we definitely want an allow list. (I originally thought otherwise, but having thought more, I'm sure you are right.)

Perhaps it should be analagous to `trackAllow` - basically a set of `trackAllow` that is like its run at the beginning of every rule? If so, `shakeTrackAllow :: [FilePattern]` seems like the best name for it and type. If we do that, then you'd expect `shakeTrackInside :: [FilePath]` rather than `shakeLintInside`, since what you are really doing is having an impact on the track mechanism, and the fact that during lint the track stuff gets actually caught is more of a detail. With those two renamings, I think we're good to go.

ndmitchell:
PS. Very keen to merge this, since I think the best way to fix the build on appveyor (https://ci.appveyor.com/project/ndmitchell/shake/build/1.0.5) is to merge this, where your track allow/ignore stuff makes it work properly.

jacereda:
I'm not sure I understand... Should I just s/`shakeLintInside`/`shakeTrackInside`/ and s/`shakeLintIgnore`/`shakeTrackAllow`/? Should I also change both to take `[FilePattern]`? Or do you mean the whole thing should be handled in `trackChange`?

ndmitchell:
I'll just merge and fix it up - just a few little changes.

PR #321, opened 2015-10-21
Fix linted paths filtering

jacereda:


ndmitchell:
Whoops! And thanks. I need to get fastrace installed on my dev machine!

jacereda:
If that's a windows box, wait until I commit some fixes I made to the windows version (tomorrow or so).

PR #322, opened 2015-10-22
Fix windows suite

jacereda:


ndmitchell:
FilePath normalisation is pretty subtle, so I'm mostly going to trust you - they seem quite sensible.

jacereda:
All fsatrace paths should be normalised (and if not, we can consider it a bug in fsatrace) and normalising all paths again seems too expensive.

ndmitchell:
Should it not be the responsibility of the trackAllow stuff to do the normalisation - so it can expect fsatrace or anything else to feed it paths any way they want, and then the tracking should normalise it. That's the way I do it for need, where you can submit any type of filepath, and I normalise it for you. I also have a fast bytestring normalise just for that purpose.

src/Development/Shake/Command.hs
```diff
@@ -207,8 +207,9 @@ fsatraceFiles file = do
     xs <- parseFSAT <$> readFileUTF8 file
     let reader (FSATRead x) = Just x; reader _ = Nothing
         writer (FSATWrite x) = Just x; writer (FSATMove x y) = Just x; writer _ = Nothing
-    frs <- liftIO $ filterM doesFileExist $ nubOrd $ map normalise $ mapMaybe reader xs
-    fws <- liftIO $ filterM doesFileExist $ nubOrd $ map normalise $ mapMaybe writer xs
+        existing f = liftIO . filterM doesFileExist . nubOrd . mapMaybe f
```
jacereda: Are we talking about the `normalise` removal in this line? The purpose at this point is to make the comparison with the paths in `shakeLintInside` robust, but it seems cheaper to normalise just the `shakeLintInside` list (probably 1 or 2 entries) than the lists of accesses.

ndmitchell: No, that's entirely fine, assuming that fsatrace generates normalised paths is very reasonable here, since we know the paths originate with fsatrace. I just don't want to assume normalisation later on, as users can call trackWrite themselves (I'm honestly not sure if we do make that assumption or not!)

PR #329, opened 2015-11-08
FSATrace with path argument

jacereda:


ndmitchell:
Given that fsatrace is a binary, any reason for not just requiring it to be on the `$PATH`?

jacereda:
State is bad and PATH is state ;-)
Not really, I made some adjustments to fsatrace and it should now be possible to just invoke it from PATH. It was assuming the full path to the executable was used to invoke it, now it will try to figure it out.

ndmitchell:
Awesome, that seems quite a bit easier to get going with.

PR #330, opened 2015-11-08
Stack complains about other-modules

jacereda:


ndmitchell:
I've also raised https://github.com/commercialhaskell/stack/issues/1323 and https://github.com/commercialhaskell/stack/issues/1324 to try and make these checks a bit better, and to have them tested automatically.

PR #333, opened 2015-11-12
Cycle through a precalculated random bools in nonDet

jacereda:
I'm getting excessive allocation/times in `nonDet` in empty builds (taking more than `getFileInfo`). Something like this seems to improve the situation.

### DIFF
--- a/src/Development/Shake/Pool.hs
+++ b/src/Development/Shake/Pool.hs
@@ -22,10 +22,11 @@ import System.Random
 -- Monad for non-deterministic (but otherwise pure) computations
 type NonDet a = IO a
 
+rnds :: [Bool]
+rnds = map (unsafePerformIO . const randomIO) [0..99]
+
 nonDet :: NonDet [Bool]
-nonDet = do bs <- unsafeInterleaveIO nonDet
-            b <- randomIO
-            return $ b:bs
+nonDet = return $ cycle rnds
 
 -- Left = deterministic list, Right = non-deterministic tree
 data Queue a = Queue [a] (Either [a] (Maybe (Tree a)))


ndmitchell:
Can you also attach the profile? What does this improve things from/to? Certainly happy to do something like this, but I'm a little wary of cycle due to CAF space leaks.

jacereda:
I'm not particularly attached to the implementation, I'm sure you'll come up with something better. I don't know if there's any leak in this approach, but the results seem to improve:

[bbs.prof.branch.txt](https://github.com/ndmitchell/shake/files/32756/bbs.prof.branch.txt)
[bbs.prof.master.txt](https://github.com/ndmitchell/shake/files/32757/bbs.prof.master.txt)
[bbs.timings.branch.txt](https://github.com/ndmitchell/shake/files/32758/bbs.timings.branch.txt)
[bbs.timings.master.txt](https://github.com/ndmitchell/shake/files/32759/bbs.timings.master.txt)

ndmitchell:
Wow, that's an impressive gain. Yes, I'll take a proper look later, certainly something needs doing here.

jacereda:
Just a reminder, if you profile something don't forget the `-j`, I forgot it once while profiling master and it kept me wondering where `nonDet` had gone for a while.

ndmitchell:
I'm not really sure what guarantee I was hoping for here - certainly enqueue and dequeue with random bools seems a little  pointless - doing the randomness twice doesn't really give you more randomness. I guess the dequeue is important so nothing gets left there too long, and the enqueue to maintain balance on the tree?

I've rewritten entirely, using a HashMap to keep balance (making enqueue deterministic), then a single Int to dequeue randomly, leaving out the nasty list of booleans and hopefully being faster than your alternative or mine. Can you please rebenchmark?

jacereda:
The overhead is still noticeable. Just how random do you need the order to be? Is using Data.Sequence and inserting randomly at the beginning or end an acceptable approach? If so, I can PR an experiment I made implementing that. The results (look at removeTree/insertTree):

[bbs.prof.master.txt](https://github.com/ndmitchell/shake/files/39618/bbs.prof.master.txt)
[bbs.prof.seq.txt](https://github.com/ndmitchell/shake/files/39624/bbs.prof.seq.txt)

ndmitchell:
Let me have another go. I had assumed Map.size was O(1), but it is O(n) - so it shows up at the top of the profile instead of being completely irrelevant. I'm not sure if randomly inserting at the beginning/end is enough or not, in fact I have disturbingly little idea what is required - the large tests I ran all used the full thing. Let's see what a size cache does.

ndmitchell:
Now I cache the size, I'd appreciate another profile run.

jacereda:
I rechecked master, it's much better now. Allocations went down from 7 to 5 GB and performance is roughly the same as my experimental branch at https://github.com/ndmitchell/shake/compare/master...jacereda:seq-pool

Perhaps the memory reduction is due to the new `parseFSAT`?

My branch gets rid of the `NonDet` monad and makes `dequeue` pure, is that desirable? Not sure how random it is though.

ndmitchell:
Yep, I suspect the memory reduction is more likely due to `parseFSAT` - I don't think this patch should change much memory usage. Although the fact that performance is roughly the same as `seq-pool` makes me happy enough to call it a day.

My best guess is that dequeue needs to be random, so that everything has multiple independent chances of being dequeued, whereas if you only enqueue randomly then something might be placed at the end and be guaranteed to be last. Given a pool which is always 100, given a random enqueue you might end up being absolute last, and be the last retrieved. Given a random dequeue you always have a 1/100 chance of getting out, so on average will always be dequeued in about 50 steps, I think. But I'm not super sure - as long as it's no longer in the profile I'll just leave it for now.

ndmitchell:
Also, changing `fsatrace` to only spit out unique results should also have had a big performance impact.

ndmitchell:
Closing, since this no longer seems to be the bottleneck. Please do raise tickets for any other things that pop up on the profile that you feel shouldn't be there.

PR #339, opened 2015-12-02
Update URLs to Ninja manual

Hodapp87:
Config.hs uses the link
http://martine.github.io/ninja/manual.html#_lexical_syntax, which goes
to the manual via a redirect to ninja-build.org, but the anchor does
not work.  This updates those links to
https://ninja-build.org/manual.html#_lexical_syntax so that they
actually point to the correct section.


Hodapp87:
Wow, that was quick :)

ndmitchell:
Thanks for the patch! I found another instance of the old website, so I've fixed that up too.

PR #341, opened 2015-12-11
Fix spacing issues when using fsatrace

jacereda:


jacereda:
Ok, I guess something like this requires implementing the escaping rules present in fsatrace. On the positive side, it would allow things like

    cmd Shell "foo.exe" ["arg 1", "arg 2"]

without worrying about escaping. Are you OK with that approach or do you want to keep compatibility and require the quoting for the args?

ndmitchell:
I consider this level of test a complete pass - I've kicked off the extra travis build just to be sure, but it looks like FileLock issues again (I'm still trying to figure out why this test is a little less robust, but every time I investigate, it starts working...). In general, consider anything with < 2 travis failures to be a success, provided the failure is not at compile time.

For the code, any reason for removing the Ninja pieces? Is that orthogonal, or is it important? I worry a lot about the > 8000 character branch, which I know I hit in a Ninja build in practice, but is not in the test quite. It's certainly a cleanup though. I guess I should probably have an explicit Ninja at > 10K chars patch.

ndmitchell:
For info, I'm out Mon-Wed, so probably won't be able to merge this before Thursday.

jacereda:
When I removed the >8000 chars code I was thinking it shouldn't be a ninja-only thing. An easy workaround would be to require sh to be present in order to support long lines. That could take the form of a `UnixShell` option that uses sh on Windows instead of cmd.

ndmitchell:
Using the Unix shell on Windows would give different semantics. I think I probably need to track down the original test case, turn it into a real test case, see if it has any effect outside Ninja, and go from there. I half-wonder if the `.bat` approach solves it anyway.

ndmitchell:
Oh, and I definitely wouldn't want to do the > 8000 approach anywhere else - it's pretty vile, if it makes a real Ninja build work then great, but I don't want anyone outside that relying on it.

jacereda:
It doesn't work when put in a .bat, I just tested a 16k echo command and it truncates the output.

jacereda:
Taking the length should be slowing down the build, but if that use case has to be supported I can bring it back. Probably the good thing would be to fix that build to use `@file` syntax in the link line (the usual failure point).

jacereda:
Well, if the length test was performed as ByteString it wouldn't slow down anything...

ndmitchell:
I'm not super sure how the Ninja > 8000 really helps - I thought the limit was that regardless. I don't think the length really slows anything down, since the list gets traversed multiple times in ways which are must slower (e.g. to concat, convert to a CString). Once you are spawning a process, everything else is cheap in comparison.

ndmitchell:
So I've reviewed, a few thoughts:

* For sh on Linux we still create a temporary file, is that necessary? And we also call `sh -- s.bat`, I wonder if some non-default shells (which people do alias to `sh`) might choke on that? Any reason not to `sh -c command`?
* Previously Shell with no Autodeps/lint went via Shell in Haskell, whereas now it always goes via our script file. Might it be better to go via Shell when we can? Or maybe the consistent semantics are preferable - I think the .bat version is probably more correct.

jacereda:
Aliases shouldn't work if we invoke `/bin/sh`. I've replaced `/bin/sh -- s.bat` with `/bin/sh -e s.bat`, if the user decides to use multiline shell scripts the `-e` should be present. Some manual pages don't mention it, but I tried with bash/csh/ksh/zsh and all of them handle it properly. Would you prefer to invoke via `sh` instead of `/bin/sh`?

jacereda:
As for going always through the script file, I thought it would avoid surprises.

I was also thinking the whole .bat thing could have been much simpler if it was instead implemented by translating `cmd Shell "xxx"` to `cmd StdIn "xxx" "sh"  "-e"`, but that would mean no `StdIn` support for `Shell`.

ndmitchell:
Apologies for not having had time to review this yet - I've been snowed under since I got back from holiday. I expect to get at it tomorrow or Monday.

ndmitchell:
Since on Linux I don't think it's ever useful to go via a file, I modified it to always run `/bin/sh -c shell`, and that works fine with `execv` and `fsatrace`. For Windows, we always go through a `.bat`, and I'm on the fence about whether that is the right thing to do or not. My guess is probably it is the right thing to do for consistency. Thanks for all your work on this!

PR #370, opened 2015-12-23
Fix putWhen

### LINKED ISSUE #369: putLoud / putQuiet do the opposite thing
Using `putLoud` with default verbosity currently displays nothing, whereas `putQuiet` does.
This is due to a bug in `putWhen`:

```
putWhen :: Verbosity -> String -> Action ()
putWhen v msg = do
    Global{..} <- Action getRO
    verb <- getVerbosity
    when (verb >= v) $ -- BUG! it should be: when (v >= verb) $
        liftIO $ globalOutput v msg
```

01mf02:
This should fix #369.

### DIFF
--- a/src/Development/Shake/Core.hs
+++ b/src/Development/Shake/Core.hs
@@ -594,7 +594,7 @@ putWhen :: Verbosity -> String -> Action ()
 putWhen v msg = do
     Global{..} <- Action getRO
     verb <- getVerbosity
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch, most appreciated! Let's continue all the discussions in #369.

ndmitchell:
Agreed in #369 that this isn't the way it was intended.

Issue #376, opened 2015-12-29
"//*.c" rule matches absolute path


ezyang:
If I write a rule matching all c files (including those in subdirectories) using the pattern `//*.c`, I in fact also match absolute paths which are NOT in my subdirectory:

```
GOOD:
Prelude Development.Shake> "//*.o" ?== "foo.o"
True
Prelude Development.Shake> "//*.o" ?== "fo/baro.o"
True
BAD:
Prelude Development.Shake> "//*.o" ?== "/fo/baro.o"
True
```
If you attempt to depend on a build product in an absolute path, your rules will spuriously match against it.

Furthermore, Shake is unclear about how needs on absolute paths are supposed to be handled in any case.

ezyang:
I have a patch.

ezyang:
PR: #377

ndmitchell:
There are a few tickets you raised, so I'll discuss the broad themes here, and then once we figure everything out here go back and update them.

So the general plan is to remove `//` and make people only use `**` - as you've seen in other tickets, `//` plays terribly with the `</>` operator. From now on, I'll discuss with `**` examples only.

The intended behaviour of `**/*.c` is meant to match _any_ `.c` file, including those with an absolute path. In general Shake can work with both absolute and relative paths just fine, it's just a matter of convention that most people are mostly building files relative to `.` and don't use absolute paths for it. If both Windows and Linux made absolute paths start with `/` I'd be open to requiring `/**` for absolute paths, but alas, that's not the Windows way at all, so would be problematic. You could potentially make `./**` work for relative paths, but it's a departure from what we currently have, and I'm a little wary of it. If `**` only matches relative paths, what would match absolute paths? `/**`?

That said, at the moment `**` has never been released in a Shake release, so we have a unique window of limited compatibility concerns, and I agree that it might not be desirable to match absolute paths, so maybe that is the way forward... I'll certainly sleep on it.

ezyang:
A few remarks:

1. I looked at the referenced systems which have `**`. For the most part, the `**` syntax descends from ant. FishEye has `**` match absolute paths; however, ant specifically states that these selectors operate on "directory trees", which means that it's impossible to match against an absolute path; the pattern is just not well-defined in that case.

[...]

PR #377, opened 2015-12-29
Make ** patterns not match absolute paths.

ezyang:

### DIFF
--- a/src/Development/Shake/FilePattern.hs
+++ b/src/Development/Shake/FilePattern.hs
@@ -58,6 +58,7 @@ data Pat = Lit String -- ^ foo
          | Star   -- ^ /*/
          | Skip -- ^ //
          | Skip1 -- ^ //, but must be at least 1 element
+         | Rel -- ^ Does not match absolute path (** = Rel, Skip)
          | Stars String [String] String -- ^ *foo*, prefix (fixed), infix floaters, suffix
                                         -- e.g. *foo*bar = Stars "" ["foo"] "bar"
             deriving (Show,Eq,Ord)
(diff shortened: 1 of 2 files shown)


ezyang:
I will admit, the walker test failure is positively baffling. I have no idea what walker is supposed to do.

ndmitchell:
Discussion at #376

ndmitchell:
Looking at the patch, I don't see it handling Windows paths, so `C:/foo` will match `**` even though it really shouldn't. I think it only makes sense for `Rel` to be at the start, and to be an immediate directory prefix test, so I might make it not part of the pattern desugaring and just do an explicit test - if `**` is a prefix then it can't be relative.

As to what relative is, I probably won't use the definition from `FilePath`, since there (on Windows) `/foo` and `C:foo` are both considered relative. In truth they aren't absolute, but they aren't fully relative either, and for this purpose I think that's "not in the current directory" so shouldn't match.

ezyang:
It looks like you'd have to make more invasive changes to make it not match absolute paths on Windows, I agree. Having `**` not match absolute paths is no longer on the hotpath for me, since I no longer use it in my build system.

ndmitchell:
Great, even now it's not on the hot path, do you still think it makes sense to do? My guess is it does.

As a matter of interest, could not comment on any tickets/pull-requests that are still on your hotpath?

ezyang:
I think it still makes sense for `**` to not match absolute paths, because then it doesn't compose nicely with `</>`.

ndmitchell:
> As a matter of interest, could not comment on any tickets/pull-requests that are still on your hotpath?

That was a typo, which almost entirely reverses the sense of the message, For clarification I meant:

As a matter of interest, could you comment on any tickets/pull-requests that are still on your hotpath?

ndmitchell:
Looking at the code, it turned out to be way easier to just test if the pattern required relative paths, and the path was relative. I didn't merge you diff directly, but did go through and take your comments/test corrections and incorporate them. Thanks for the pull request, and sorry it took so disastrously long to get merged!

PR #379, opened 2015-12-29
use MultiWayIf in Core.hs

Heather:


ndmitchell:
Thanks for the patch! However, I still support GHC 7.4, which doesn't yet have MultiWayIf, so it's not appropriate for Shake.

While I am a fan of a concise if syntax (and often use `case () of`), I'm not sure it's as useful in this particular instance because in some ways I really want the `-- condition 1` etc to stand out on its own on a line - it's almost "code" in itself.

For info, on line 289 the extra "redundant" brackets around the type signature are intentional - they are used to suggest that it takes one argument and returns a function, so should be partially applied when used.

I hope you're having a happy holiday season, and am grateful for the patch, even though not merged.

PR #384, opened 2015-12-30
Implement shakeExtra, which is a TypeRep to Dynamic map in ShakeConfig.

### LINKED ISSUE #383: Generalize 'ShakeOptions' argument in Rule's storedValue method
It would be very helpful if 'ShakeOptions' had a user-programmable, 'TypeRep' map which they could use to add their own data to be available when implementing 'storedValue' and 'equalValue'. Otherwise, it's too difficult to implement a Rule instance when the 'storedValue' needs some (non-global) configuration. File gets a config flag; why not our own rules too!

ezyang:
Fixes #383.

### DIFF
--- a/src/Development/Shake/Types.hs
+++ b/src/Development/Shake/Types.hs
@@ -8,6 +8,9 @@ module Development.Shake.Types(
 
 import Data.Data
 import Data.List
+import Data.Dynamic
+import Data.HashMap.Strict (HashMap)
+import qualified Data.HashMap.Strict as HashMap
 import Development.Shake.Progress
 import Development.Shake.FilePattern
 import qualified Data.ByteString.Char8 as BS
@@ -76,8 +79,8 @@ data Change
 --
 --   @ 'shakeOptions'{'shakeThreads'=4, 'shakeReport'=[\"report.html\"]} @
 --
---   The 'Data' instance for this type reports the 'shakeProgress' and 'shakeOutput' fields as having the abstract type 'Function',
---   because 'Data' cannot be defined for functions.
+--   The 'Data' instance for this type reports the 'shakeProgress' and 'shakeOutput' fields as having the abstract type 'Hidden',
+--   because 'Data' cannot be defined for functions or 'TypeRep's.
 data ShakeOptions = ShakeOptions
     {shakeFiles :: FilePath
         -- ^ Defaults to @.shake@. The directory used for storing Shake metadata files.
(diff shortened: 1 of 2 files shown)


ndmitchell:
Why not use the existing unordered containers package, which is already a dependency?

src/Development/Shake/Types.hs
```diff
@@ -164,24 +173,27 @@ shakeOptions = ShakeOptions
...
-    ,"shakeLiveFiles","shakeVersionIgnore","shakeProgress", "shakeOutput"]
+    ,"shakeLiveFiles","shakeVersionIgnore","shakeProgress", "shakeOutput", "shakeExtra"]
 tyShakeOptions = mkDataType "Development.Shake.Types.ShakeOptions" [conShakeOptions]
 conShakeOptions = mkConstr tyShakeOptions "ShakeOptions" fieldsShakeOptions Prefix
-unhide x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 x16 x17 x18 x19 x20 x21 y1 y2 =
-    ShakeOptions x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 x16 x17 x18 x19 x20 x21 (fromFunction y1) (fromFunction y2)
+unhide x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 x16 x17 x18 x19 x20 x21 y1 y2 y3 =
+    ShakeOptions x1 x2 x3 x4 x5 x6 x7 x8 x9 x10 x11 x12 x13 x14 x15 x16 x17 x18 x19 x20 x21 (fromFunction y1) (fromFunction y2) (fromFunction y3)
+    -- NB: use of fromFunction for y3 is a little bit of a hack; but
+    -- arguably we should call this fromHide not fromFunction
```
ndmitchell: I'm happy renaming Function to Hide - seems entirely reasonable.

ndmitchell:
Thanks for the patch. Looks good!

ezyang:
TypeRep only has an Ord instance, no hashable instance, so we need an Ord-based container.

I wonder why Travis is failing.

ndmitchell:
Probably failing due to https://github.com/ndmitchell/shake/blob/master/src/Test/Self.hs#L87 - that list needs containers adding.

However, I'm not super happy about adding an entire containers dependency just for a a single map. Also, TypeRep does have a Hashable, because I asked for it, or are you thinking of it somehow else?

ndmitchell:
Note the Hashable TypeRep instance is in Data.Hashable, so its not orphan, but if you just have TypeRep in scope at ghci, you won't see it.

ezyang:
Oh I see! Yes, I was looking at the Haddock docs for TypeRep which of course aren't going to report a TypeRep hashable. OK I'll switch it to an unordered container.

ezyang:
I don't know what's going on with the resource busy error in 7.4. Intermittent maybe?

ndmitchell:
Probably, I wouldn't overly worry - GHC 7.4 has some segfaults and race conditions. Shake seems to be hitting this more and more on Travis (perhaps upgraded Travis machines?) so I might add test retrying on older versions.

Issue #348, opened 2015-12-14
Development.Shake.Rule unclear about stored value semantics; does "not stored externally" mean phony?


ezyang:
The manual states:

> For rules whose values are not stored externally, storedValue should return Just with a sentinel value and equalValue should always return EqualCheap for that sentinel.

I don't understand what this means. For example, if I want to make a separate namespace of phony targets (rather than the built-in phony support, which lives in the same namespace as files), this is not something that is "stored externally"; so should I follow the advice here? The answer is no!

```
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MultiParamTypeClasses #-}
import Development.Shake
import Development.Shake.Classes
import Development.Shake.Rule
import Control.Monad

newtype Phony = Phony String
    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)
newtype PhonyVal = PhonyVal ()
    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)

instance Rule Phony PhonyVal where
    storedValue _ _ = return (Just (PhonyVal ()))
    equalValue _ _ _ _ = EqualCheap

main :: IO ()
main = shakeArgs shakeOptions $ do
    action (void (apply [Phony "foo", Phony "bar"] :: Action [PhonyVal]))
    rule $ \(Phony x) -> Just (liftIO (putStrLn x) >> return (PhonyVal ()))
```

Then I get:

```
ezyang@sabre:~/Dev/labs$ ./aaa 
bar
foo
Build completed in 0:01m
ezyang@sabre:~/Dev/labs$ ./aaa 
Build completed in 0:01m
```

which is not what phony rules are supposed to do. So it seems stored value should always return `Nothing`. (What should equalValue do? I don't see a difference either way.)

(For that matter, what is something that is not stored externally?!)

ndmitchell:
`equalValue` is doc bug, I don't remember writing that piece of the docs, and it's definitely not true. If your type supports sensible equality (as `()` does), then you don't need to worry.

The basic idea behind storedValue is you are asking "is the value consistent on disk". If you have a value on disk, you should return it, and if it sometimes has a value but not always, Nothing. However, if your rule never has a value on disk then returning Just () says "never rebuild" unless one of the other dependencies changes, and Nothing says "always rebuild". I'll make the docs much clearer, and probably add an example or two.

[...]

Issue #191, opened 2014-11-12
Idea: "task" rules that are half-way between file rules and phony rules


3noch:
File-based rules run when dependencies change or when the file is missing. Phony rules don't produce a file and always run. I would like to have something in the middle: a rule that does not produce a file, but only runs when its dependencies change. Of course, this behavior can be achieved with temp-files. But there is overhead to get that working properly, especially if you want these "task" based rules to match FilePatterns just like file-based rules do.

This might be a bad idea or it might not fit with the goals of the project. That's ok. I just wanted to propose it in case it was useful. I've achieved this with temporary files and it's working well.

ezyang:
Suppose rule A depends on task B. If you don't care about rerunning rule A if task B ran this round, you can implement tasks today in Shake.

```
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE MultiParamTypeClasses #-}
import Development.Shake
import Development.Shake.Rule
import Development.Shake.Classes

newtype Task = Task String
    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)

instance Rule Task () where
    storedValue _ _ = return (Just ())
    equalValue _ _ _ _ = EqualCheap

main :: IO ()
main = shakeArgs shakeOptions $ do
    action $ (apply1 (Task "foo") :: Action ())
    rule $ \(Task k) -> Just $ do
        need [k]
        putNormal k
```

Example:

```
ezyang@sabre:~$ runghc Test.hs 
Build completed in 0:01m
ezyang@sabre:~$ touch foo
ezyang@sabre:~$ runghc Test.hs 
foo
Build completed in 0:01m
```

One downside of this approach is that Shake's unchanging rebuilds will assume that any rule which depends on the task does NOT need to be rebuilt.

If you want to rerun any of the rules that depended on your task when the task reruns, then you can't use a trivial sentinel in your rule: you need something like a timestamp (a monotonically increasing counter), and you need to save it somewhere. One way to do this is to associate tasks with files and look at the timestamp in the file (rm to force the task to rerun): that's the tempfile approach suggested above.

[...]

Issue #185, opened 2014-10-24
Investigate oracle/cache relationship


ndmitchell:
Discussing Oracle and Cache, it's clear there are multiple overlapping concerns. Effectively you want:
- A way of caching computation, sometimes transiently.
- A way of splitting a single value into multiple values.
- A way of storing something in the Shake database.

Currently `newCache` gives you transient computation, and `addOracle` gives you a way of storing in the database, plus a bit of splitting, but also brings `alwaysRerun` along for the ride - which is sometimes awful.

Maybe the type of addOracle and newCache combined looks something like:

```
addOracle' :: Action (q -> Action a) -> Rules ()
newCache' :: Action (q -> Action a) -> Rules (q -> Action a)
```

There's certainly something `Action (q -> Action a)` to `q -> Action a` in caching for splitting that probably wants surfacing a bit more directly. Using different names `new` vs `add` is probably a mistake as well.

ezyang:
Correction: the output of oracles IS transiently cached (and persistently cached), since they are rules, and rules are only one once an invocation of Shake!

My thought was that everything should be a rule. I thus spent some time trying to work out how to encode caches as rules. I came to the conclusion that what we are missing is a few new states in the build system: `LazyLoaded`, `LazyChecking` and `LazyReady`. They form a graph like this:

```
LazyLoaded -> LazyChecking
LazyChecking -> Building
LazyChecking -> LazyReady
LazyReady -> Building
```

The basic idea is that a `LazyLoaded` rule is halfway between a `Missing` and a `Loaded` rule: if we ever actually directly NEED the value (via another Building key), we will need to build this key, but if no one ever actually directly needs this value, AND we otherwise could have skipped rebuilding the rule (because none of its dependencies changed), then it will always stay in the `LazyReady` stage and never actually get built.

This is kind of complicated, so I recommend just keeping caches, and maybe just supporting a more "rule-like" interface for them; i.e. instead of `newCache`, have `addCache` and `askCache`. `addCache` has slightly different semantics from `newCache`, in that you wouldn't be allowed to cache two different functions of the same type (that would make askCache different) but it would help you avoid having to "plumb" the cached functions around (the big benefit of the rule interface, despite the fact that it's terribly ambiguous.)

[...]

Issue #382, opened 2015-12-30
No way to support build flag like `-dynamic-too` (Shake considers change in file pattern grounds for rebuild)


ezyang:
GHC supports a flag named `-dynamic-too` with the following semantics:

1. Without `-dynamic-too`, GHC outputs an `.hi` file and an `.o` file when you compile an `.hs` file.
2. With `-dynamic-too`, GHC outputs the same as (1) (bit-for-bit, in fact), AS WELL AS `.dyn_hi` and `.dyn_o`.

I attempted to implement this in Shake by writing a conditional on the `-dynamic-too` flag, providing a `&%>` rule which matched against just `hi` and `o` without the flag, and `hi`, `o`, `dyn_hi` and `dyn_o` with the flag.

However, after implementing this, I ran into mysterious spurious recompilation problems.

The problem is that Shake treats a `&%>` rule as defining a "multiple file rule", with the other rules depending on it. So, without `-dynamic-too`, Shake generates a rule named "*.hi *.o", and with it, it generates a rule "*.hi *.o *.dyn_hi *.dyn_o". If I go from passing `-dynamic-too` to not passing it, the "*.hi *.o" rule doesn't exist so Shake rebuilds it.

@ndmitchell , you've mentioned that you're planning on changing how multi-file rules are implemented. I don't know if this solves this problem. Within the current framework, I don't really know how to solve this.

ezyang:
Using Shake's internal API, I implemented a working version of `-dynamic-too`: https://github.com/ezyang/ghc-shake/blob/master/GhcShake.hs#L670

It's hella janky, but here's the basic idea: I define a new rule key which essentially is a 'Module' (plus some extra info). From this module, I can compute all of the output file paths that will show up if I build it: if I'm building module A, then normally I will get A.hi and A.o, unless `-dynamic-too` is on, then I'll also get A.dyn_hi and A.dyn_o.

The answer type consists of a `Maybe FileA` for every possible output file we can output. `storedValue` loads up `FileA` in much the same way `Files` is implement, but the equality test does something fun: depending on what mode we're in (`-dynamic-too` or not), it will check more or less of the data structure in question. Thus, without `-dynamic-too`, an out-of-date `dyn_hi` file will not cause the validity check to fail.

[...]

PR #389, opened 2015-12-31
Implement storedValueE which supports 'NeverRecomp'.

ezyang:
This commit introduces a new method to 'Rule', 'storedValueE',
whose return type is extended to be 'MaybeStored', a data type
that can reports 'AlwaysRecomp' (previously 'Nothing'),
'RecompIfNotEq' (previously 'Just') and 'NeverRecomp' (new!)
'NeverRecomp' causes Shake to treat the value stored for the
rule in the database as canonical; the rule will only be compiled
if (1) there was no value in the database to begin with, or
(2) dependencies changed.

See discussion in #191 (and #185).  This architectural extension
makes it possible to write oracle rules which do NOT always
rerun, although I haven't added any porcelain for that.

Signed-off-by: Edward Z. Yang ezyang@cs.stanford.edu

### DIFF
--- a/src/Development/Shake/Core.hs
+++ b/src/Development/Shake/Core.hs
@@ -147,7 +147,7 @@ class (ShakeValue key, ShakeValue value) => Rule key value where
     storedValueE opts k
         = maybe AlwaysRecomp RecompIfNotEq `fmap` storedValue opts k
 
-    -- | Like 'storeValueE', but this method does not have the ability
+    -- | Like 'storedValueE', but this method does not have the ability
     -- to return 'NeverRecomp'.  For backwards-compatibility.
     storedValue :: ShakeOptions -> key -> IO (Maybe value)
     storedValue opts k = do
--- a/src/Development/Shake/Core.hs
+++ b/src/Development/Shake/Core.hs
@@ -72,13 +72,15 @@ import Prelude
 -- * A way to compare two states of the same individual artifact, with 'equalValue' returning either
 --   'EqualCheap' or 'NotEqual'.
 --
--- * A way to query the current state of an artifact, with 'storedValue' returning the current state,
---   or 'Nothing' if there is no current state (e.g. the file does not exist).
+-- * A way to query the current state of an artifact, with 'storedValueE' returning the current state 'RecompIfNotEq',
+--   or 'AlwaysRecomp' if there is no current state (e.g. the file does not exist).
 --
 --   Checking if an artifact needs to be built consists of comparing two @value@s
 --   of the same @key@ with 'equalValue'. The first value is obtained by applying
---   'storedValue' to the @key@ and the second is the value stored in the build
---   database after the last successful build.
+--   'storedValueE' to the @key@ and the second is the value stored in the build
+--   database after the last successful build.  If you return
+--   'NeverRecomp' from 'storedValueE', then this check is always
+--   assumed to succeed.
 --
 --   As an example, below is a simplified rule for building files, where files are identified
 --   by a 'FilePath' and their state is identified by a hash of their contents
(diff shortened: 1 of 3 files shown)


ezyang:
This patch should also fix #348 doc problem.

ndmitchell:
Seems sensible - and making the result an ADT rather than a Maybe is probably the right way to go even if we didn't add the new alternative. I'll have to sit down and review naming tonight and the rest in more detail, since it is quite key.

ndmitchell:
Thinking further, I'm increasingly convinced this is what I should have done all along, and a lot of the oracle/rule complexity comes from missing this option.

ezyang:
Yeah, happy to paint the naming bikeshed whatever color you want! (NeverRecomp is not the best name because we CAN still recompile... if the dependencies change.)

ndmitchell:
I have started on the New Year's bubbly already, so I won't be merging something so sensitive tonight, and my next few days are all childcare related, so I'll probably sit on this until middle of next week. I still think it's a fantastic idea though.

ezyang:
Here's a prototype of a "persistent cache" which is saved over Shake runs:

```
{-# LANGUAGE MultiParamTypeClasses, GeneralizedNewtypeDeriving, DeriveDataTypeable, ScopedTypeVariables, ConstraintKinds #-}
{-# LANGUAGE UndecidableInstances #-}

module PersistentCache(
    addPersistentCache, askPersistentCache
    ) where

import Development.Shake
import Development.Shake.Rule
import Development.Shake.Classes
import Control.Applicative
import Prelude


newtype CacheQ question = CacheQ question
    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)
newtype CacheA answer = CacheA answer
    deriving (Show, Typeable, Eq, Hashable, Binary, NFData)

instance (ShakeValue q, ShakeValue a) => Rule (CacheQ q) (CacheA a) where
    storedValueE _ _ = return NeverRecomp


-- | A persistent cache is a function from question type @q@, to an answer type @a@,
--   which is cached across runs to Shake.  Cached values are not
--   recomputed unless any of their dependencies change.
--
--   Persistent caches are used similarly to oracles, but unlike
--   oracles, they are not rerun every invocation of Shake.  Unlike
--   'newCache', these caches ARE saved to disk (and thus the value
--   must be serializable), and you are not allowed to have two
addPersistentCache :: (ShakeValue q, ShakeValue a) => (q -> Action a) -> Rules (q -> Action a)
addPersistentCache act = do
    rule $ \(CacheQ q) -> Just $ CacheA <$> act q
    return askPersistentCache


-- | Get information from a cached 'addPersistentCache'. The
-- question/answer types must match those provided to
-- 'addPersistentCache'.
askPersistentCache :: (ShakeValue q, ShakeValue a) => q -> Action a
askPersistentCache question = do CacheA answer <- apply1 $ CacheQ question; return answer
```

ndmitchell:
Replying this after an excessive delay (sorry about that), but not yet enough time to actually merge/hack it. My current best guess is that the types/names should be:

    data StoredValue
        = StoredValue a     -- ^ I have a value persistently stored (e.g. on disk)
        | StoredMissing     -- ^ I was expecting a value persistently stored (e.g. on disk) but didn't find it
        | StoredPrevious    -- ^ I didn't expect a value

    storedValueEx :: ShakeOptions -> key -> IO (StoredValue value)

Note that I'm making a rule talk about what it has, rather than what it would like future stages to do (e.g. StoredPrevious vs DontRecompile). I'm still debating in my head if that is the right thing to do.

For the `addPersistentCache` my thought is actually to unify all Oracles and have `addOraclePersist` and `addOracleRecompute`. `addOraclePersist` (or whatever it gets called) would use `StoredPrevious`, while `addOracleCompute` would just be sugar over `addOraclePersist` plus a call to `alwaysRerun`.

Thinking about this issue, I was forced to delve into the reason for having `storedValue` at all. For example, you could have all rules always do `StoredPrevious` and then `alwaysRerun` would be a primitive. File rules would be implemented by having two types of rules, UserFile and TimestampFile. UserFile would be like the current one, but it would also add a dependency on TimestampFile. The TimestampFile would be a single rule that did the modtime check currently in `storedValue`. This neatly unifies oracles and rules, and makes it clear that you can't use a rule without registering it (as the Binary instance requires). However, it has some disadvantages. First, we might have to store the file name twice with different keys (although there might be an interning scheme here). Second, storedValue is done single-threaded without the thread pool overhead. Does the single-threadedness hurt? (Perhaps not for filetime). What would be the thread pool overhead?

However, I suspect I'll merge this, perhaps with a renaming, and then think about the future direction in a separate ticket.

ndmitchell:
Another important part of storedValue is doing the clever avoiding computing the hash by having a lazy thunk. Harder if we have evaluated rules in the DB.

ezyang:
If you can work out the performance implications, I like the proposed unification. I think it would help with #382 since now I can write my rules as depending on the timestamp file to build a custom multi-file rule (since I no longer need to jam it in `storedValue`). But it seems to me that you're probably better of treating always rerun rules specially, since they are now perf critical.

ndmitchell:
Yes, if I managed to unify the two pieces, then you'd have "normal" rules (which would look a look like oracles) and would be `PreviousValue`. Then you'd have a single special distinguished rule, likely baked into the core data structures as a Bool which just sets a flag on a rule, that would be always rerun (and we'd also drop all other dependencies on always rerun since they aren't interesting) - so once always rerun is the only special thing we can optimise it a lot more.

ndmitchell:
I thought this through. The flaw is that the calculate the timestamp of a file rule must run twice, once before (to decide whether we should run or not), and once after (to capture the value that was built). That's fundamentally different from a rule that can only run once. It may mean we need a "quick" rule associated with a type, as well as a fallback rule, or some other formulation.

ndmitchell:
I think the right surface syntax for this is `addOracleRerun` and `addOracleCache`, where the existing oracle stuff is `addOracleRerun`.

ndmitchell:
Note that a change to adjust StoredValue to an ADT is present in the pull request #416.

ndmitchell:
I think this and the discussions around it lead to the `RunChanged` type and the new `addOracleCache`, so I'll close this with implemented in other ways.

PR #390, opened 2015-12-31
Documentation patches

ezyang:
I feel a bit bad for spamming the bug tracker with so many issues, so here are some documentation patches. I'll push more patches to this PR as I write more.


ndmitchell:
Thanks! Don't worry too much about the doc tests - if the docs are good I can easily fix those up when I merge - it's fairly confusing code which I want to refactor/redo anyway so makes sense that I suffer.

ndmitchell:
Do you want me to merge now? Or merge once you've reached a documentation fixed point?

Don't consider the previous lack of doc writing to be a sign they were starting to overwhelm me - more a reflection that this is a busy time of year, so I'm stacking them all up to do whatever is left in January.

src/Development/Shake/Core.hs
```diff
@@ -65,7 +65,34 @@ import Prelude
...
+--   allowed for type classes and not allowed in Shake either!)
+--
+-- * 'Eq' and 'Hashable' are used on keys in order to build hash maps
+--   from keys to values.  'Eq' is used on values to test if the value
+--   has changed or not (this is used to support unchanging rebuilds,
+--   where Shake can avoid rerunning rules if it runs a dependency,
+--   but it turns out that no changes occurred.)  The 'Hashable'
+--   instances are *strictly* used only at runtime (and not
+--   serialized to disk), so they do not have to be stable across runs.
+--   TODO: What is Hashable on values used for?
```
ndmitchell: Hashable on Value is of no use, it's there for consistency. It's not difficult to implement if you have Binary, so consistency is easier than preciseness.

ndmitchell:
Thanks! All merged, and I cleaned up a few of the remaining questions in the docs.

PR #403, opened 2016-01-08
Fix stdcall warnings on x64

jacereda:


jacereda:
I'm not sure this is the right way to deal with this.

ndmitchell:
I can't think of a better way, so my inclination is to merge this and consider it fixed. Looks good to me.

jacereda:
Shouldn't the appveyor script be changed to do something like "stack --resolver=nightly-xxx"?

jacereda:
Or maybe just add a stack.yaml...

ndmitchell:
There is no nightly with the new version of extra - I think it should have been produced last night but it's still missing. I expected one day broken, but it's a bit annoying to have two.

ndmitchell:
I ideally don't want to add a stack.yml because it requires maintaining, which is a bit annoying, especially if using the nightly is always expected to work.

jacereda:
I tried with nightly-2016-01-08 and extra-1.4.3 is there.

ndmitchell:
Ah, we aren't failing to find a build plan since nightly would work. Instead we have managed to corrupt the appveyor cache meaning the build plan has a `.git` file but no contents, which Stack then fails to update properly. Annoying, and I fixed it in Stack recently, so hopefully temporary.

ndmitchell:
Hmm, turned out we needed something in the .hs files otherwise it doesn't work in ghci or if we runhaskell or ghc --make it (as we do in the test suite). Since the .hs files have to have an `#ifndef` they might as well define the variable themselves - so I moved it there instead of in the cabal. As it turns out we have 3 hs files that require it, instead of 3 instances in the Cabal file, so doesn't make too much difference.

jacereda:
I guess ghc needs an `apicall`to do the right thing in each platform or something like that.

ndmitchell:
That would be very useful!

PR #406, opened 2016-01-14
Filter out non-existing files (temporaries) reported by Autodeps

jacereda:


ndmitchell:
Thanks, seems a sensible idea.

jacereda:
The alternative would be to report deletes and interpret the result, but this seems simpler.

Issue #262, opened 2015-06-20
Regularise the test suite


ndmitchell:
Currently, every test entry does entirely different things with the arguments. I should probably say `!` prefixed arguments are special and go to the args,  `$` prefixed is for obj, and unprefixed ones are phony requests.

ndmitchell:
General thoughts about how to move forward:

* Move to shakenCwd everywhere
* Remove the non-cwd branches
* Remove the obj argument (always set to id)
* Make the want implicit, only pass on special arguments via --test (perhaps retrieved via an oracle, use withTest?)

ndmitchell:
There is now a single interface.

Issue #192, opened 2014-11-12
Allow a single key type to map to multiple value types.


ndmitchell:
At the moment Phony rules, FileHash rules and ModTime rules all get mushed into one, because you want all of them to match on `String` keys. If some variant of `apply_` was introduced, which didn't care about the result type (as most of them don't) you could use separate rules for each set of rules but make them all depend on the same key type.

ndmitchell:
Note that also rules currently do two modtime checks per file. With apply_ I could make that one and shrink the database a bit.

ezyang:
If you don't need an open universe on strings, you could use a GADT to encode the return type. This has the added benefit that you can still support apply with a return type (pattern matching on the GADT to determine the return type). This is especially true since you don't want to allow overlaps for rules which have different outputs but the same key.

For that matter, if you apply for a specific value, and there exists a rule for that key, but a different value, what are you supposed to do?

ndmitchell:
I definitely want an open universe, since then all the Shake rule types can be entirely separate from each other, without overlap.

I don't think there is any problem with different rules returning different keys, or overlaps. If given a particular string there are two matches, then that's an error, just like now. Overlaps of different types are not special.

> For that matter, if you apply for a specific value, and there exists a rule for that key, but a different value, what are you supposed to do?

Not totally sure what that is asking. When you call `apply` the first time it will find the associated action from `rule` and run it, producing an output type. If you do it again, it won't be looking at the rule/action, it will just see a key/value pair in the database, with types that must be unique (given the key type and the value type), and then check it. I guess the thing that stops any problems is that each entry in the database is indexed by key, key type, and value type. (Or, if it's not, it will have to be.)

ezyang:
Yes, you'll have to index by key type and value type. If you only know about the monomorphic interface, that's kind of surprising, but OK!

ndmitchell:
What do you mean by the monomorphic interface - via `need`? I don't think such users are even aware that the underlying rule returns a value.

ezyang:
Yeah. (Well, it's only confusing if you decide you need to define some sort of rules on not filenames, and then you discover that it's indexed by key type and value type.)

ezyang:
I was rereading this ticket and I am now confused by https://github.com/ndmitchell/shake/issues/192#issuecomment-88927977

How does multiple output values help with the `&%>` problem? As far as I can tell, you still need a rule key that represents the set of files your pattern is matching over.

ndmitchell:
[...]

PR #416, opened 2016-02-08
Misc. changes

Mathnerd314:
This does 5-ish things, all interspersed:
- Switch Travis to container infrastructure (no benefit other than faster queuing AFAICT; the caching doesn't seem to add any speedup)
- Change `depends` from `[[Id]]` to `[Id]` (not sure why it was `[[Id]]` in the first place)
- Change Maybe to StoredValue, as in https://github.com/ndmitchell/shake/pull/389#issuecomment-168140073
- General refactoring and test-fixing
- Some initial work on making .Forward part of Shake Core (currently I'm trying to make `ForwardQ Key` work)

Feel free to cherry-pick as you see fit.


ndmitchell:
Thanks for the changes! A lot to pick through, but a few quick remarks/questions from what I've seen:

* For Travis faster queuing is a bit nice, but the file looks far more complex. If I do decide to go that route I'd push the complexity into my Neil tool (https://github.com/ndmitchell/neil) and have much less in Shake - that way all my projects can share it. How much faster is the queuing? Is there some expiry date on the old stuff?

* `[[Id]]` is very important, specifically to https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Database.hs#L312. If a rule does `need [a]; b <- ... readFile' a; need [b]` then you have to check if `a` needs rebuilding before you try speculatively rebuilding `b`, as the use of `b` may no longer be correct if `a` has changed. As a result I try rebuild checking everything in each pile of `[Id]` in parallel, but do the whole `[[Id]]` in series.

* For `StoredValue` I'll do this as part of #389 at the same time, and hope to get to that tonight ideally.

* What's the advantage of making `Forward` part of `Core`? At the moment it's deliberately off to the side since it doesn't rely on being intertwined with `Core` and provides a very different end-user experience.

* What's the advantage of moving `ShakeValue` into `Classes`? Or do you think that just looks cleaner? (I'm somewhat inclined to agree, but wonder if there is more to it than that.)

* For the Command and Docs test fixes, was there some circumstance in which they broke before? I haven't seen such failures in my testing, including on Travis/Appveyor, so would be keen to know if my testing is incomplete or something else is going on.

ndmitchell:
I've now pushed a comment about `[[Id]]`, and a test that checks the observable behaviour that comes about as a result. Certainly something so important and subtle deserves a failing test if it breaks.

Mathnerd314:
* Well, the old stuff took a minute or two to queue, the containers take a few seconds. And I'm not sure it's slower; it could be faster (particularly if ninja is cached). You can delete the cache manually (from Travis) and from within the script but otherwise it doesn't expire AFAIK. 
* I see. I get "ab" on the test, which I guess means that both rules are run.
* OK. I look forward to seeing what you do.
* The ``cachedAction`` primitive that Forward provides is more general than ``newCache``, I've been experimenting further if you look at the commits I pushed just now.
* The Command test was broken on my system when I started; perhaps because '.' isn't in my path.
* The Docs test was broken by exporting ShakeOptions from Rule; I'm not sure whether anything else would break it

ndmitchell:
* Sounds like the Travis improvements are a good idea then. I'll probably take the code you've done and shove it back into the Neil project and branch it out from there.

* The command test fixes make sense, thanks.

* I guess the docs test fix follows one of your previous changes? that makes sense too.

* So for the `cacheAction`, it seems like you want that to be available in the Core so you can implement Shake.Forward on top of it. Is that motivated by a particular use of cacheAction in a normal build system? Or are you trying to clean up the Forward stuff to avoid global IORefs?

Mathnerd314:
Well, the goal is similar to @ezyang's in #382; I want to write a "forward-thinking" build system. basically to run a command and specify its input/output files, like so:
```
newtype BuildModule = BuildModule InputFile deriving (ShakeValue)
foo :: Rules ()
foo = 
    rule $ (\BuildModule f) -> do
        (ghc, path) <- need (GHC, f)
        listOfOutputFiles <- cmd ghc f
        provide [map toFile listOfOutputFiles] -- borrowed from Pluto
    action (BuildModule "MyFile.hs")
```
Shake doesn't do anything like that at present, except in ``cacheAction``, from what I've seen.

ndmitchell:
OK, I've started merging bits and pieces. There's a lot here, so I'm just trying to separate out the small fixes from the bigger pieces. Notes:

* I've merged the ShakeValue stuff. Thanks.

* The change to the Command test to account for `.` not being on `$PATH` was actually a bug in the test. That executable was meant to be located in a subdir, so that I could test things like changing the current directory and ensuring I still picked it up properly. That implicitly relied on the assumption that `obj "" /= getCurrentDirectory`, which is exactly what I started to change in #262. I've now fixed the test by pushing the test executable inside a directory, which hopefully fixes it for you too.

* You have a patch that exports more things from Development.Shake.Rule, specifically things that are already exported from Development.Shake.Classes or Development.Shake. Any reason? My thought was to keep that to just the minimum "extra" things that people can only get from there, almost like an `Internal` module (but with more API stability and an expectation for people to use it for their own rules). Partly that's to have an easy to draw line (otherwise you keep wanting more and more things in there), and partly to keep the documentation short (only things you didn't know about before).

ndmitchell:
I experimented with the Travis scripts. I converted my `neil` template project over, so you can see the revised files at: https://github.com/ndmitchell/neil. Generally for the "shared infrastructure" between all my projects (e.g. the Travis script), I prototype it in `neil` and then share the code via that, so all my projects are doing the same Travis stuff in the same way. I am happy to take patches there, or for you to do it against Shake and I'll move the concepts over, as I did in this case.

I suspect caching makes no difference because you cached the `~/.cabal` files, but missed the package config files which are in `~/.ghc`. As a result the packages were there, but GHC/Cabal didn't know they were there, and ended up rebuilding. I implemented that (by simply caching both, and not worrying about invaliding it etc), and saved the 100s that the `neil` project typically spends installing dependencies. Definitely a win, if we move to the containers. Looking at the Shake timings, it would save 71s, so nice, but not earth shattering. In contrast, on Hoogle it's 569s and HLint it's 306s, so very useful for those projects, and when done generically in `neil`, it's "free" for Shake.

Unfortunately, there is a bit of a snag. On GHC 7.8.4 and later the builds take vastly longer than they did before. Looking at a dashboard at https://travis-ci.org/ndmitchell/neil/builds/108992902 I see GHC 7.4 and 7.6 take 2 minutes (about what the old system took), while the others are closer to 8 minutes. Doing a build and looking at it ticking, it seems the GHC compilation of the package in question goes vastly slower, but the dependent packages still go about the same speed. I have no idea why. Any thoughts? If that could be fixed I'd be ready to move to the new containers.

ndmitchell:
Flipping sudo back to true fixes the speed problems, but obviously then removes the ability to cache and fast queuing which would be useful. I've raised a Travis ticket: https://github.com/travis-ci/travis-ci/issues/5623

ndmitchell:
I'll deal with the introduction of an ADT in #389. Other than that, are there are small pieces you think should be taken from here? Or are we just left with the forward dependencies?

Assuming we're only left with the forward dependencies, I'm still not quite on the same page. There are forward driven build systems (e.g. [Fabricate](https://code.google.com/archive/p/fabricate/)), and they have forward execution of everything. They essentially provide a script which runs in order, with caching of previously executed steps. I believe Development.Shake.Forward in repo provides a reasonable approximation of that. Is there anything you feel is missing?

There are also backward driven build systems, which start at the things you want to produce and run rules to get to the leaves. Shake is one such example, but so is Make etc.

I'm not sure if it's possible to combine forward and backward in one coherent build system which provides access to the features of both. In Shake you can certainly pick either, but currently both would probably have issues.

I see that Pluto has `provide`, but in their paper they mention the word `provide` 67 times, but only a handful of those refer to their `provide` function. I'm still lost as to how you implement it, what happens if the result is incoherent (e.g. two people claim to have provided the same thing), if you can run subsets of the build system to produce certain targets (I guess not without being able to statically evaluate the provides?). Do you have intuitions for these questions? (Even if not the same answers that Pluto arrives at, or pointers to the Pluto paper section if I just overlooked it.)

Assuming provide must not overlap (and that you must check that yourself), and that provide doesn't cause rules to be looked up, wouldn't `needed` work in place of `provide` in your example?

Mathnerd314:
> exporting more things from Development.Shake.Rule

My thought was to make every instance of Rule import from Development.Shake.Rule instead of .Core, which perhaps was overly ambitious. Re-exporting ``.Classes`` still seems fine to me (it adds 1 line to the haddocks and is very convenient) but perhaps the others should go away again.

> Travis container building takes longer

Interesting; I have no explanation either. I'll limit my request to explicitly setting sudo: true in ``.travis.yml``.

>  Are we just left with the forward dependencies?

There's a lot of refactoring between Core/Database. The stored / equal of Rule could be replaced with ``analyse :: key -> value -> IO (AnalysisResult value)``; do you have an opinion on that?

> forward/backward distinction

Forward dependencies are more than just running a script. For example, ``cabal install js-flot`` produces ``.cabal/lib/x86_64-linux-ghc-7.10.3/js-flot-0.8.3-8Jf4CdemM2vD7PZsKaF5pA``; you would never ask for that long hash directly, as is required in Shake's backward build system.

> Is there anything you feel is missing?

I'd like the ability to make separate namespaces besides files; even phony rules seem like a bit of a hack at present.

> Pluto has provide

They also use the term "generated files", which is more informative in searching. The build algorithm is on [page 98](www.informatik.uni-marburg.de/~seba/publications/pluto-incremental-build.pdf#page=10). Comparing ``needed`` file dependencies with generated files, it seems that they have 2 extra features: they store a reference to the builder/rule that called them, and then throw a fatal error if another rule tries to provide them or if another rule tries to require/need them without first depending on the generating rule.

ndmitchell:
> Re-exporting .Classes still seems fine to me (it adds 1 line to the haddocks and is very convenient)

Yep, seems reasonable, and also reexporting it from Development.Shake. Or perhaps, only Development.Shake? Having it in both is a tad unnecessary if Rule is just the "extra" bits. But practically, if it's in Development.Shake, then it's available to anyone defining rules anyway.

> Interesting; I have no explanation either. I'll limit my request to explicitly setting sudo: true in .travis.yml.

I'm not aware of any weaknesses in the `.travis.yml`, other than lack of caching and slow startup, but both of those are inevitable with `sudo:true`. My hope is Travis will explain how to fix `sudo:true` and then I'll be able to incorporate your nice improvements.

> The stored / equal of Rule could be replaced with analyse :: key -> value -> IO (AnalysisResult value);

I have been increasingly coming around to that point of view, and even to the idea that dropping the Rule class entirely is probably the ultimate direction. My description of AnalysisResult is quite different to yours though, I'll try and finish my thoughts and write them down somewhere, and then we can figure why/how they differ, and if either can be enhanced by the other.

> For example, cabal install js-flot produces .cabal/lib/x86_64-linux-ghc-7.10.3/js-flot-0.8.3-8Jf4CdemM2vD7PZsKaF5pA; you would never ask for that long hash directly

Indeed. In a backward build system I guess you'd have a node that represents `js-flot-0.8.3` and that would generate a stub, plus the hash version. I'll have a think and try and describe exactly the pattern I would encourage for that. It may turn out to be roughly what you are imagining. Even if not, writing down my way of faking it seems valuable.

> I'd like the ability to make separate namespaces besides files

Each key lives in a separate namespace already. Or is the point that files and other string-like keys live in the same namespace but are different? I think #192 might be what I think of as a solution to this.

> Pluto provide stuff

I think `needed` plus lint checking does everything you describe. Having `provide` as a separate function to be semantically clearer might be valuable even if it's an alias for `needed`. Having `provide` do the little bit of the lint check relevant to this file would be good though, and mean errors got thrown at the right point. Thanks for the links, I'll follow them.

ndmitchell:
Looking into it, the reason I don't reexport all the type classes by default is mostly because Binary contains a `get` method and thus clashes with the State monad. I'd rather my code didn't clash with the state monad by default. Given that information, I think not exporting is the right choice - I've noted that in the Classes module (I suspect I've rediscovered this more than once).

> they have 2 extra features: they store a reference to the builder/rule that called them, and then throw a fatal error if another rule tries to provide them or if another rule tries to require/need them without first depending on the generating rule.

Checking into the details, `needed` does indeed provide that. Tweaks that could be made: a `provides` (I think the pluralisation seems right here) function to make the semantic intent clearer; `needed` currently only checks if `shakeLint` is turned on, that seems wrong; error messages need expressing in terms of `provides` not `needed`; if you generate twice it says it changed but not built twice (although the lint feature does do that if we `trackWrite`); `provides` should call `trackWrite`.

I note that the Pluto check contains a weakness in that if you never require the rule that calls `provides` then you never know that another rule missed calling it. That seems a fundamental limitation of any build system that doesn't rebuild everything and has opaque function (e.g. EDSL), so I don't see how they or Shake can do any better.

Thinking about it, I wonder if lint should be the default? Catching difficult errors is good, and I could check how high the performance penalty really is.

Mathnerd314:
> I note that the Pluto check contains a weakness in that if you never require the rule that calls provides then you never know that another rule missed calling it

If you remove generated files from version control, then clean builds of the rule will fail because the file doesn't exist. It's then a matter of policy as to what's considered "generated" and occasionally doing clean builds (or FSATrace) to catch stray files.

ndmitchell:
@Mathnerd314 - good point, that does seem to answer most of the problems - I don't think it's a problem in practice.

Mathnerd314:
> Each key lives in a separate namespace already. Or is the point that files and other string-like keys live in the same namespace but are different? I think #192 might be what I think of as a solution to this.

``apply_`` is impossible to implement. My plan is
```
newtype PhonyQ = Phony String
newtype FileQ = <current file type>
```
etc. (newtypes for every namespace), and a function ``String -> Maybe Key`` in the Rule class for parsing command line arguments.

ndmitchell:
Nice idea, I think I prefer that to my formulation...

ndmitchell:
I'm now starting to think the original `apply_` formulation is better... But I'm definitely wavering, trying to understand all the details.

I'm sorry it's taking so long to process all the things you've come up with - I do appreciate all the pieces, and it makes me glad to know that someone manages to understand the internals (significantly reduces the bus factor of Shake). For info, I've got a paper deadline Tuesday, then another one 16th March, but hope to fit in some of the backlog between then and all of it afterwards.

ndmitchell:
Just so you're are of my current thoughts, I wanted to get 0.15.6 out (now done) and then as part of 0.16 I intend to break the Rules instance, which gives us more freedom to incorporate some of these changes. I'll be working on that over the next few days - expect a ticket with my proposed outline to go up shortly (I'll mention you in it).

Mathnerd314:
Sounds fine. I'm watching the repository though, no need to @mention me.

ndmitchell:
I think all the pieces were upstream, with the exception of the `Forward` changes. For that, you have generalised `cacheAction` to be polymorphic, and relied on `registerWitness`. Inspired by your code, I changed how it worked so it stores the underlying ByteString and only decodes when a user asks for it, which means I don't need to registerWitness. The downside is it's probably less efficient, but if this approach is popular we can figure out how to make it efficient (I have a few ideas).

PR #439, opened 2016-04-20
Change some dashes into proper Unicode en-dashes

iblech:
Feel free to close the pull request unmerged! I like those Unicode en-dashes, but I know not everybody does.


ndmitchell:
I like the visual display of endash, but dislike it being in the source code, because some systems will choke on it and it's very difficult to visually detect the difference by eye.

As an alternative, how about using ` -- ` in the Markdown, then changing the website generator to put in `&endash;`? The place in the website generator is https://github.com/ndmitchell/shake/blob/master/website/Main.hs#L115, and it should be something like:

    reformat mode code (TagText x:xs) = TagText (replace " -- " " &endash; " x) : reformat mode code xs

iblech:
Good idea! I'll look into it in the next couple days.

ndmitchell:
Did you make any progress on this issue? Even just inserting the double dashes instead of emdashes? If you did, I'll merge it in, if not, I'll probably merge this request then use that as a guide for where to use --.

iblech:
Thanks for reminding me of this issue and sorry for not replying sooner, it fell under my radar. I'm now preparing the updated pull request.: -)

ndmitchell:
Awesome! Many thanks (and please don't feel any pressure that you _must_ do the request - I'm both grateful for code and ideas).

iblech:
Sure, didn't feel any pressure. :-)

I believe I'm done with the pull request now. Your code proposal worked just fine, with one modification: One cannot use `&endash;`, since the `&` sign will be encoded as `&amp;` by other layers of the code. The code now uses `\x2013`, that is, it insrts a literal Unicode en-dash into the output stream.

Please tell me if I you want me to squash the two commits into one.

ndmitchell:
Thanks! Github now has an option to squash two commits into one when you merge, which I did (I wouldn't usually, but if I don't I just _know_ that one day I'll have a bisect that lands between these points, and then some tool I never knew I had will choke on the non-Ascii characters!).

ndmitchell:
In fact this patch immediately broke my generation, because it tried to write out non-Ascii characters, and that's not allowed in my default locale. I fixed that, but also modified the code to replace the dashes with their HTML equivalent. I'll regenerate the website later - hopefully tonight.

ndmitchell:
I regenerated the web page, looks good. Thanks once again.

iblech:
Aha! I didn't notice that during my testing, because (as a German needing umlauts) I have an UTF-8 locale. I quite like your postprocessing filter!

PR #444, opened 2016-04-28
Refactored copyFileChanged to extract whenFilesDifer

jacereda:
I also removed what seems to be a leftover `withTempFile`.
Note that `copyFileChanged` isn't exactly equivalent since it won't perform the `need` when the file doesn't exist. Is that a problem?

### DIFF
--- a/src/Development/Shake.hs
+++ b/src/Development/Shake.hs
@@ -111,6 +111,7 @@ module Development.Shake(
     -- * Explicit parallelism
     parallel, forP, par,
     -- * Utility functions
+    whenFilesDifer,
     copyFile', copyFileChanged,
     readFile', readFileLines,
     writeFile', writeFileLines, writeFileChanged,
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -44,7 +44,7 @@ import Development.Shake.FilePath
 import Development.Shake.FilePattern
 import Development.Shake.Types
 import Development.Shake.Rules.File
-import Development.Shake.Derived(writeFile', withTempDir)
+import Development.Shake.Derived(writeFile', withTempDir, withTempFile)
 
 ---------------------------------------------------------------------
 -- ACTUAL EXECUTION
@@ -559,12 +559,6 @@ instance Arg a => Arg (Maybe a) where arg = maybe [] arg
 
 ---------------------------------------------------------------------
 -- UTILITIES
-
--- Copied from Derived. Once Derived no longer exports cmd stuff, import from there.
-withTempFile :: (FilePath -> Action a) -> Action a
-withTempFile act = do
-    (file, del) <- liftIO newTempFile
-    act file `actionFinally` del
 
 -- A better version of showCommandForUser, which doesn't escape so much on Windows
 showCommandForUser2 :: FilePath -> [String] -> String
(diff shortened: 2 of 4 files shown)


jacereda:
Would extra be a better place for `whenFilesDifer`?

ndmitchell:
The left over `withTempFile` is a good call.

Rather than have `whenFileDiffer` why not have `fileEq :: FilePath -> FilePath -> IO Bool`? That returns True only if both files exist and have the same contents. Adding that to `extra` seems very reasonable. You can then do `whenM (fileEq file1 file2) $ ...` for your `whenFileDiffer`, which avoids the `MonadIO` issue (currently `extra` deliberately doesn't depend on `transformers`). You could also switch to Lazy byte strings, which would give you O(1) memory usage.

I think `copyFileChanged` should definitely still call `need`, since that's an important part of what it does. I don't think implementing the `need` additionally is hard using the `fileEq` approach (or using `whenFileDiffer` either).

jacereda:
Sounds much better. Is it ok to make `extra` depend on `bytestring`?

ndmitchell:
Ideally not... Writing them directly on top of `System.IO.hGetBuf` shouldn't be too hard though and avoids the dependency while also making the IO more explicit.

ndmitchell:
If you have any issues writing `fileEq` do let me know - I'm happy to write it myself if that helps. Looking into the details, you probably also want to copy the definition of memcmp out of ByteString (it's only a few lines): http://hackage.haskell.org/package/bytestring-0.9.1.4/docs/src/Data-ByteString-Internal.html#memcmp

I've copied the bit removing the duplicate withTempFile and pushed it, so closing the remainder of this ticket and we can do the `fileEq` thing in the extra package, then use it here.

jacereda:
Let me try to write the `fileEq`, it will be my first attempt at dealing with raw pointers in haskell...

jacereda:
I was trying to measure the speed of my `fileEq` using `hGetBuf` versus the `ByteString.Lazy` variant with a large file (500Mb) but the second fails miserably:

    fileEqLBS :: FilePath -> FilePath -> IO Bool
    fileEqLBS p1 p2 =
      doesFileExist p1 &&^ doesFileExist p2
      &&^ (withBinaryFile p1 ReadMode $ \h1 ->
             withBinaryFile p2 ReadMode $ \h2 ->
             liftM2 (==) (hFileSize h1) (hFileSize h2)
             &&^ liftM2 (==) (LBS.hGetContents h1) (LBS.hGetContents h2))

This is the error I get:

    hGetBufSome: illegal operation (handle is closed)

What am I doing wrong?

jacereda:
Ok, I'll answer myself:

```
fileEqLBS :: FilePath -> FilePath -> IO Bool
fileEqLBS p1 p2 =
  doesFileExist p1 &&^ doesFileExist p2
  &&^ (withBinaryFile p1 ReadMode $ \h1 ->
         withBinaryFile p2 ReadMode $ \h2 -> do
           c1 <- LBS.hGetContents h1
           c2 <- LBS.hGetContents h2
           return $! c1 == c2)

```

The strictness is needed to ensure the comparison is made while the handles remain open. Some year I'll get used to laziness...

jacereda:
Looks like I got something faster than the above with `hGetBuf`. I'll send the PR.

ndmitchell:
All merged and released as extra-1.4.4.

ndmitchell:
I switched copyFileChanged to use fileEq, given the performance improvements you implemented.

PR #456, opened 2016-05-25
Use displayException for prettier errors when possible

sid-kap:
Apparently this function has been a part of the `Exception` typeclass [since base 4.8](https://hackage.haskell.org/package/base-4.9.0.0/docs/Control-Exception-Base.html#v:displayException).


ndmitchell:
Great idea! I try to keep CPP out of Shake where I can, so I've just added displayException to the extra library in https://github.com/ndmitchell/extra/commit/dc237ee9bc39fbbe8bdd561b7a396ec7b90b4fb7. Tonight I'll release a new version of extra, then use that and remove the CPP here.

sid-kap:
Thanks!

ndmitchell:
I now use extra to provide displayException in a version compatible way.

Issue #431, opened 2016-03-12
Add a test for dependencies on directories


ndmitchell:
Sometimes people think dependencies can be on directories, and do things equivalent to:

```
"my/dir" %> \out -> createDirectoryIfMissing True out
```

I should improve the error message and ensure it really is an error in the test suite. Driven by #430.

ndmitchell:
Turns out everything around directories works fine with the modtime checks, which is a complete bug. I've added a currently disabled test case, but will fix it and then we'll have proper errors.

ndmitchell:
Added on both Windows and Linux.

PR #457, opened 2016-05-25
More descriptive error when trying to `need` a directory

sid-kap:
Related to #431

I was getting this error, and it took me a while to understand the cause. Hopefully a more detailed error will make it easier for users.

(In my case, it was especially confusing because I was editing legacy code that was abusing `need` to generate directories. When I upgraded by repo to use the current master version of Shake, I started getting this error.)

### DIFF
--- a/src/Development/Shake/FileInfo.hs
+++ b/src/Development/Shake/FileInfo.hs
@@ -108,7 +108,7 @@ getFileInfo x = BS.useAsCString (unpackU_ x) $ \file ->
         let peek = do
                 code <- peekFileAttributes fad
                 if testBit code 4 then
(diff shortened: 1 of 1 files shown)


ndmitchell:
That's a much better error message, thanks! I've improved it a bit more, and also wired it through the Shake.Errors module, which provides better formating and consistent style for all errors - I should really have been using that all along.

Given how useful this feature is to users upgrading, I'll aim to make a release tonight including it.

sid-kap:
Thanks!

ndmitchell:
I made the 0.15.7 release containing this feature, hopefully it will help those who come after you.

PR #460, opened 2016-05-30
Update demo and manual to use Stack GHC

bhipple:
Since the docs recommend installing Shake with stack, a raw command to
ghc will likely fail due to missing libraries if a user tries to re-run
the Shake demo after the initial `stack exec -- shake --demo` cmd.

### DIFF
--- a/docs/Demo.md
+++ b/docs/Demo.md
@@ -5,6 +5,7 @@ To get started with Shake, you need to:
 * Install the [Haskell Stack](http://haskellstack.org/).
 * Type `stack install shake`
 * Run `stack exec -- shake --demo`, which should produce something similar to the trace below.
+* To repeat the build after the initial demo, run `stack exec ./build.sh` in the demo directory.
 
 <pre>
 <!-- nosyntax --><b>% Welcome to the Shake v1.0 demo mode!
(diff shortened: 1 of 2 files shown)


docs/manual/build.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/sh
 mkdir -p _shake
-ghc --make Build.hs -rtsopts -with-rtsopts=-I0 -outputdir=_shake -o _shake/build && _shake/build "$@"
+command -v stack >/dev/null 2>&1
+if [ $? -eq 0 ]; then
```
ndmitchell: I tend to use `if type stack > /dev/null 2>&1; then` as the guard here. Any reason to prefer your formulation or mine? (I'm not a big shell scripter, so defer to your knowledge here!)

ndmitchell:
Thanks! Although I slightly wonder if the guidance should be run to `stack exec -- build` and keep the stack bit outside of Shake? I would rather Shake didn't have to pick a side in the Cabal vs Stack thing, but maybe it's just something I can't avoid?

bhipple:
That's a good point; keeping the `build.sh` script as simple as possible seems preferable.  

Regarding `type`, it'll most likely work everywhere, but [Shellcheck](https://github.com/koalaman/shellcheck/wiki/SC2039) points out that it's not [POSIX standard](http://unix-linux.todaysummary.com/q_unix-linux-programming_99198.html).

ndmitchell:
Ah, interesting about ShellCheck - I haven't run it on that particular script for a while.

Looks good, but on Windows it doesn't work - https://github.com/commercialhaskell/stack/issues/2225 - I think this is a Stack bug so we'll see what they say before going forward. If it looks like it will be difficult to solve on their side we can have different Windows/Linux instructions, but I'd rather avoid that if we can.

ndmitchell:
Stack have now fixed so `stack exec build` works on Windows. All merged and switched to `build.bat` on Windows (mostly for better symmetry with `build.sh`). Since the code to run `build.bat` is only in the dev version of Stack I'll hold of regenerating the website right now and leave it a little bit. Thanks for the patch!

PR #489, opened 2016-10-08
Fix haddock syntax so that linking works and spurious / is gone

benclifford:


ndmitchell:
Thanks for the patch.

My understanding was that `\/` was the correct way to write `/` in Haddock, since `/` means "italics", so you have to escape the `/`. Which Haddock did you test with? Things like `Monoid/Monad` appearing with spaces around the `/` would be less correct from an English sense. Where is the spurious `/`? Does it occur in all the instances you fixed, or is one of them the one that made you investigate this

benclifford:
It appears in hackage, for example I initially noticed it here: https://hackage.haskell.org/package/shake-0.15.10/docs/Development-Shake-Command.html#t:Process which I took to be some canonical proof of this being broken syntax.

Without the quotes, it renders as quote name slash otherslash quote name slash, and with the patch it renders as hyperlinked-name space slash space hyperlinked-name.


Once I investigated that, I was using haddock from local ghc-7.10.3 .deb install (I think):
```
Haddock version 2.16.1, (c) Simon Marlow 2006
Ported to use the GHC API by David Waern 2006-2008
```
and it occurred in all of the instances that I fixed (one of which is rendered by `cabal haddock --internal` so doesn't appear in the hackage rendering)

ndmitchell:
Thanks for the effort tracking this down. In Haddock `/` means italics, so you need to write `\/`, but in the special case of only having one `/` on a line, it is ignored. As a result, I've gone through and kept your spaces and added `\/`, which hopefully keeps Haddock happy.

PR #493, opened 2016-10-19
Correct help docs for single letter flags

### LINKED ISSUE #492: DOC: -j=N should be -jN
Shake's help message states `-j=N` but that doesn't work, instead `-jN` does.

traytonwhite:
For single letter flags with optional args, remove `equal` sign from
help documentation

Fixes #492


ndmitchell:
Thanks! Good spot, I appreciate the fix. I also noticed that the `--help` output has the same issue (I originally copied some of the webpage from there) so I've fixed that too.

PR #500, opened 2016-11-09
Fix section headers in Development.Shake haddock

chris-martin:
At https://hackage.haskell.org/package/shake-0.15.10/docs/Development-Shake.html the section headers in the module documentation aren't integrated into the rest of the table of contents, and they render like

```
== WRITING A BUILD SYSTEM ==============================
```

This fixes that and makes the HTML docs look a bit nicer.


ndmitchell:
That's way nicer - thanks a lot for the patch!

PR #503, opened 2016-12-15
Implement getShakeExtra

bgamari:
This is a convenient utility to extract a particular key from shakeExtra.
Unfortunately we can't make this interface typesafe since shakeExtra is a bit
too general: there is no known relationship between the key TypeRep and the
value at that key.

### DIFF
--- a/src/Development/Shake/Internal/Core/Action.hs
+++ b/src/Development/Shake/Internal/Core/Action.hs
@@ -3,7 +3,7 @@
 
 module Development.Shake.Internal.Core.Action(
     runAction, actionOnException, actionFinally,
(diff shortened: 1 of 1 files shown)


ndmitchell:
I think the right thing to do is to add a corresponding `addShakeExtra` which does the insert correctly, and also at startup check all items in shakeExtra obey the types/values invariant. I'll add those changes on top of this.

ndmitchell:
I added `addShakeExtra` and made it an extra for `shakeExtra` to not obey the obvious invariant - meaning the failure in `getShakeExtra` is now impossible to hit.

PR #515, opened 2017-02-16
cmd_

### LINKED ISSUE #234: cmd ergonomics: making own cmd derivations
Currently it's not easy to define a derivation of `cmd` without losing its special property -- vararg-ness.
And indeed, some derivations don't have an obvious way of making them.

Exhibit one: ..to avoid specifying the `unit $` prefix before each cmd invocation:
```haskell
type a :-> t = a
cmdn ∷ S.CmdArguments args => args :-> Action r
cmdn = unit . S.cmd
```
..fails type checking.

mikeplus64:
Fixes #234

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -45,6 +45,7 @@ import Development.Shake.Internal.FilePattern
 import Development.Shake.Internal.Options
 import Development.Shake.Internal.Rules.File
 import Development.Shake.Internal.Derived
+import Development.Shake.Internal.Unit
 
 ---------------------------------------------------------------------
 -- ACTUAL EXECUTION
@@ -560,10 +561,6 @@ type a :-> t = a
 -- @
 cmd :: CmdArguments args => args :-> Action r
 cmd = cmdArguments []
-
-class Unit a
-instance {-# OVERLAPPING #-} Unit b => Unit (a -> b)
-instance {-# OVERLAPPABLE #-} a ~ () => Unit (m a)
 
 -- | See 'cmd'. Same as 'cmd' except with a unit result.
 -- 'cmd' is to 'cmd_' as 'command' is to 'command_'.
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleInstances, GADTs, TypeSynonymInstances, TypeOperators, ScopedTypeVariables, MultiParamTypeClasses, NamedFieldPuns #-}
+{-# LANGUAGE FlexibleInstances, TypeSynonymInstances, TypeOperators, ScopedTypeVariables, NamedFieldPuns #-}
 
 -- | This module provides functions for calling command line programs, primarily
 --   'command' and 'cmd'. As a simple example:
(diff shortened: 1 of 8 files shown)


src/Development/Shake/Command.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleInstances, TypeSynonymInstances, TypeOperators, ScopedTypeVariables, NamedFieldPuns #-}
+{-# LANGUAGE FlexibleInstances, GADTs, TypeSynonymInstances, TypeOperators, ScopedTypeVariables, MultiParamTypeClasses, NamedFieldPuns #-}
```
ndmitchell: I don't see a use of either GADTs or MPTC in the code below - are they necessary?

src/Development/Shake/Command.hs
```diff
@@ -561,6 +561,15 @@ type a :-> t = a
 cmd :: CmdArguments args => args :-> Action r
 cmd = cmdArguments []
 
+class Unit a
+instance {-# OVERLAPPING #-} Unit b => Unit (a -> b)
+instance {-# OVERLAPPABLE #-} a ~ () => Unit (m a)
```
ndmitchell: The `OVERLAPPING` and `OVERLAPPABLE` are not available until GHC 7.10, and Shake still supports back to GHC 7.4. I think you can tweak it to be a language pragma on those versions - may require CPP though.

ndmitchell:
Awesome! I didn't know this was possible, A few things to clean up, but a great addition!

src/Development/Shake/Command.hs
mikeplus64: MPTC no, whoops. GADTs (or TypeFamilies and probably a few other extensions) allow the `a ~ ()` constraint.

src/Development/Shake/Command.hs
mikeplus64: Drat -- you're right. New commits split Unit into its own module that imports yet another module based on which GHC version is available. [apparently GHC doesn't use cpp to parse language pragmas so something like
```
{-# LANGUAGE CPP #-}
#if foo
{-# LANGUAGE OverlappingInstances
#endif
...
```
didn't work]

ndmitchell:
Awesome - looks great. I'll need to play with it a bit this weekend to check it really is matching my expectation, but a fantastic addition. Shame GHC makes it so hard!

mikeplus64:
I tried again with just having Unit defined in the one module and it worked -- maybe I made a typo or something earlier. But I haven't tested it with GHC 7.{4,8}, hopefully the CI can illuminate any problems ...

ndmitchell:
All works nicely - many thanks!

With your revised code I had a shot at merging the Unit module into Command and it worked, so I guess whatever you managed to fix worked there too.

mikeplus64:
Yay!

I don't feel particularly strongly on it, but I think that having a separate module for Unit is better so that *all* the typeclass instances in Command.hs aren't marked as overlapping in GHC <7.10. I can only imagine it being a problem if an instance head in Command.hs is written incorrectly, though it still does seem like Better DesignⓇ to me.

ndmitchell:
Good point. I'm not so worried about that since GHC 7.8 is mostly legacy by now, and if they did overlap, GHC 8.0 would catch them anyway. If it had to be that way in newer compiler versions I'd have gone the separate module route.

PR #518, opened 2017-03-17
Add a '--no-colo[u]r' flag.

### LINKED ISSUE #49: Make --color a shakeOption
I would like to enable colouring by default using `shakeOptions`; it is not exposed there yet.

Also: Colours are great, they should be on by default! A bit of green makes your build system feel much friendlier.

23Skidoo:
Fixes #49.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
+/TAGS
 /*.nix
 /dist/
+/dist-newstyle/
 /output/
 /.hpc/
 /misc/win7-taskbar/Taskbar.suo
--- a/src/Development/Shake/Internal/Args.hs
+++ b/src/Development/Shake/Internal/Args.hs
@@ -118,6 +118,9 @@ shakeArgsWith baseOpts userOptions rules = do
                                                   shakeLintInside oshakeOpts
                                ,shakeLintIgnore = map toStandard $
                                                   shakeLintIgnore oshakeOpts
+                               ,shakeOutput     = if shakeColor oshakeOpts
+                                                  then outputColor (shakeOutput oshakeOpts)
+                                                  else shakeOutput oshakeOpts
                                }
 
     errs <- return $ errs ++ flagsError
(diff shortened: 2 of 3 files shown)


ndmitchell:
Thanks! I'll take a look and properly review probably tomorrow night.

ndmitchell:
This patch defaults color to True, which might be a problem on Windows or piping to a file. There are lots of trade-offs here, but the one HLint uses is to default to False on Windows, and on Linux ask hSupportsANSI on stdout. Not sure if that makes sense for Shake or not...

23Skidoo:
> This patch defaults color to True, 

I believe that this is the case even without this patch: https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Internal/Args.hs#L262

I can look into making it work like HLint does.

ndmitchell:
Sorry for taking forever to reply...

This issue is a bit more complex than feels right. There are a few issues in my mind - adding dependencies, the interplay of shakeOutput/shakeColor and whether we want to do more than just color entire lines. After thinking, I got to the point where:

* Get confirmation that ansi-terminal can be used and isn't going to cause issues for the GHC build system. Done as https://github.com/snowleopard/hadrian/issues/399, it's not an issue.
* Get ansi-terminal have support for taking a String containing ANSI escape sequences and "replaying" it in a compatible manner. Requested in https://github.com/feuerbach/ansi-terminal/issues/26.
* Make shakeOutput default to being more than putStrLn, but interpreting escape codes and replaying them on Windows.
* If shakeColor is set then pass escape codes to shakeOutput.

I think that provides the best of all worlds, and gives us the features you want. Only issue is that in the transition it makes sense for us to default to color=False, and once the features arrive in ansi-terminal, then change the default. I'm fairly certain that the default is currently false. The `yes` refers to being available in `ShakeOptions` or not, rather than anything else.

ndmitchell:
ansi-terminal doesn't want to add the code, so plan is:

* Merge this request
* Default color to False
* Write the ansi-terminal emulation for Windows
* Default color to True

23Skidoo:
>  The `yes` refers to being available in ShakeOptions or not, rather than anything else.

I can of course be wrong, but AFAICR my reason for writing this patch was that `--colour` was on by default and there was no way to disable it.

Issue #514, opened 2017-01-28
`cmdArguments` and `Arg(..)` are not exposed.


NorfairKing:
I would like to make some helper functions that allow for the usage of `Path` and `Path.IO` so I would like to instantiate `Arg` for my own data type `data PathCmdOption = forall r. PathCwd (Path r Dir)`

ndmitchell:
Do you need `cmdArguments` exposed? Or just `Arg(..)`? If exposing I'd be tempted to rejig it and expose:

```
data CmdArgument
instance Monoid CmdArgument

class IsCmdArgument a where
     toCmdArgument :: a -> CmdArgument
instance IsCmdArgument String
instance IsCmdArgument [String]
instance IsCmdArgument CmdOption
instance IsCmdArgument [CmdOption]
instance IsCmdArgument a => IsCmdArgument (Maybe a)
```

Would that be enough for you?

NorfairKing:
I think so. That would allow me to define my own command arguments, which is what I'm looking for.

NorfairKing:
Any progress here?

ndmitchell:
Not recently - I hope to circle back to Shake hacking in a few days time. It's a fairly easy change, so I hope it will be in the next release.

ndmitchell:
Maybe having `CmdOption` opaque is just too much hassle - being honest and saying it's `([CmdOption],[String])` might give more flexibility for other people to reuse the same mechanism.

NorfairKing:
I'll leave it up to you. My use-case is to be able to use a `Path` from [`path`](https://hackage.haskell.org/package/path) as a command argument.

ndmitchell:
Done as part of #526.

PR #526, opened 2017-05-04
issues 516 and 522

fadilmokhchane:
Please find attached some simple bug fixes. 
Please let me know if there is any particular issue you want me to tackle next ?

Thx Fadil


src/Development/Shake/Internal/Derived.hs
```diff
@@ -72,22 +75,30 @@ addShakeExtra x = Map.insert (typeOf x) (toDyn x)
 
 -- | @copyFile' old new@ copies the existing file from @old@ to @new@.
 --   The @old@ file will be tracked as a dependency.
+--   Also creates the new directory if necessary.
 copyFile' :: FilePath -> FilePath -> Action ()
 copyFile' old new = do
     need [old]
     putLoud $ "Copying from " ++ old ++ " to " ++ new
+    createDirIfMissing new
```
kirbyfan64: Maybe I'm wrong (and I usually am), but given `copyFile' "a" "b"`, wouldn't this end up creating the directory `b` and then copying `a` into it, resulting in `b/a` instead of `b`?

fadilmokhchane: `createDirIfMissing b`
`copyfile a b`
I haven't changed the copyfile which stills copy a to b. I added a check prior to the copyfile to ensure that the directory of b exists, if it doesn't it creates it.
No worries, all comments are welcome.

src/Development/Shake/Internal/Derived.hs
```diff
@@ -72,22 +75,30 @@ addShakeExtra x = Map.insert (typeOf x) (toDyn x)
...
     liftIO $ copyFile old new
 
+createDirIfMissing :: FilePath -> Action()
+createDirIfMissing new =
+   liftIO $ createDirectoryIfMissing True $
+          (if isWindows then FW.takeDirectory else FP.takeDirectory) new
```
ndmitchell: No need for the isWindows test - if you just use System.FilePath.takeDirectory then it is basically an alias for what you wrote and will work on both platforms.

src/Development/Shake/Internal/Derived.hs
ndmitchell: @kirbyfan64 - note that createDirIfMissing is a helper that creates the directory of the file, rather than `new` itself. Given that in the other commit I suggest removing most of the logic of `createDirIfMissing`, I'd recommend inlining whatever is left, as it will make it more clear what it is doing.

ndmitchell:
If you want to go hunting after this, #514 isn't a bad place to go.

fadilmokhchane:
I don't understand how my commits make Travis fail on a test saying Ninja is faster than Shake? Any clues? Thx

fadilmokhchane:
It's okay the tests passed with your travis changes on my fork. c.f. https://github.com/fadilmokhchane/shake/pull/2#partial-pull-merging

ndmitchell:
Thanks for all the code, looks good and all merged!

Regarding the Travis failures, I have a check that Shake must be faster than Ninja to compile Ninja (where Ninja is the fastest build system out there, apart from Shake). Unfortunately the Travis tests run on a shared box, so I get a few failures now and then as load changes over time. Generally I'm always enhancing my checking scripts, which keeps causing failures, but merging from HEAD is sufficient (although I'm usually happy to fix those up myself).

PR #529, opened 2017-06-20
Add test on file that can't exist

cblp:
If file doesn't exist, `getFileInfo` must return `Nothing`.

Why I need this: I want to build executable `myexe` and run tests with phony `myexe/test`.

### DIFF
--- a/shake.cabal
+++ b/shake.cabal
@@ -379,6 +379,7 @@ test-suite shake-test
         Test.Directory
         Test.Docs
         Test.Errors
+        Test.Existence
         Test.FileLock
         Test.FilePath
         Test.FilePattern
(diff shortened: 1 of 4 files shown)


cblp:
The only test fail is

> shake-test: ASSERTION FAILED: Bad throttling, expected to take 1.4s + computation time (cap of 0.2s), took 1.80605816s

What does it mean?

ndmitchell:
Thanks for the patch. You can probably ignore the timing error - I check things happen within N seconds on the Travis servers, and sometimes they have a big hiccup - it's a rare failure that probably isn't anything to do with you.

Regarding the patch, under what circumstances does ioeGetErrorType e == InappropriateType occur? I guess if you have `foo/bar` but `foo` is a file? If so, nice catch!

cblp:
> Regarding the patch, under what circumstances does ioeGetErrorType e == InappropriateType occur? I guess if you have foo/bar but foo is a file?

Yes. Theoretically, it may be caused by other reasons, but I can't see a problem.

Issue #502, opened 2016-11-28
Recommended way to implement quicker action alternatives


expipiplus1:
I have a build system which has a rule for producing `foo.out` from `code.in` and `data.in`. The process of compiling `foo.out` when `code.in` has changed, `slowCompile`, is very lengthy, however there does exist a process, `fastUpdate`, to *update* `foo.out` with a new `data.in` fairly quickly.

If `code.in` has changed then `slowCompile` must be run, using the new `code.in` and the existing `data.in`.

If `code.in` is unchanged but `data.in` is changed then `fastUpdate` can be run to produce a new `foo.out` from the old `foo.out` and the new `data.in`.

This is certainly possible in Shake with manual checks of the 'newness' of `data.in` and `code.in`. I'm wondering if there's a way of using the Shake machinery to do this though.

ndmitchell:
Hmm, not that I can think of - the "difficult" bit is that you only want to run one of `fastUpdate` or `slowCompile`, never both. I'll have a further think.

expipiplus1:
If Shake could expose something so an `Action` could find out why it's running then this might be nicer to do.

matthiasgoergens:
I have a slightly similar use case with bulk-updates:

I have a command that makes targets [a, b] and commands to make [a] and [b] on their own.

If we need both [a] and [b] the bulk update is faster, but otherwise it's better not to update [a] when only [b] needs change and vice versa.

(My concrete use-cases are for having to install some dependency packages with apt-get and for some code-generation script that has a long initialisation time.)

ndmitchell:
Thinking more, the general mechanism seems to be that you are writing a single rule that can produce multiple result files, and you want `needHasChanged` to ask if the `need` has changed since you last asked for it. You could imagine:

```
needHasChanged :: [FilePath] -> Action [Bool]
```

Certainly for advanced usage only, but I can see the merits. Unlike other `need` calls, this one is caller-context sensitive, so if two rules make the same `needHasChanged` call only one might report that it has changed, if both rules got run at different times.

expipiplus1:
So there would be one action which branches between the slow and fast versions depending on the result of `needHashChanged`? If any of the slow causing inputs have changed then it does the slow thing, otherwise it does the fast thing.

expipiplus1:
[...]

PR #539, opened 2017-10-06
 #502, implemented needHasChanged

erwo42:
Tried to implement `needHasChanged`.

Please review.

I tried to change as little as possible and get the desired behavior of `needHasChanged`.

### DIFF
--- a/src/Development/Shake/Internal/Rules/File.hs
+++ b/src/Development/Shake/Internal/Rules/File.hs
@@ -2,7 +2,7 @@
 {-# LANGUAGE ViewPatterns, RecordWildCards, FlexibleInstances, TypeFamilies #-}
 
 module Development.Shake.Internal.Rules.File(
-    need, needBS, needed, neededBS, want,
+    need, needHasChanged, needBS, needed, neededBS, want,
     trackRead, trackWrite, trackAllow,
     defaultRuleFile,
     (%>), (|%>), (?>), phony, (~>), phonys,
@@ -48,7 +48,7 @@ infix 1 %>, ?>, |%>, ~>
 ---------------------------------------------------------------------
 -- TYPES
 
-type instance RuleResult FileQ = Maybe FileA
+type instance RuleResult FileQ = FileR
 
 -- | The unique key we use to index File rules, to avoid name clashes.
 newtype FileQ = FileQ {fromFileQ :: FileName}
@@ -57,6 +57,12 @@ newtype FileQ = FileQ {fromFileQ :: FileName}
 -- | Raw information about a file.
 data FileA = FileA {-# UNPACK #-} !ModTime {-# UNPACK #-} !FileSize FileHash
     deriving (Typeable,Eq)
+
+-- | Result of a File rule, may contain raw file information and wether the rule did run this build
+data FileR = FileR { fileA :: (Maybe FileA)
+                   , hasChanged :: Bool
+                   }
+    deriving (Typeable, Eq)
 
 -- | The types of file rule that occur.
 data Mode
@@ -87,15 +93,28 @@ instance BinaryEx [FileQ] where
 instance Hashable FileA where
     hashWithSalt salt (FileA a b c) = hashWithSalt salt a `xor` hashWithSalt salt b `xor` hashWithSalt salt c
 
+instance Hashable FileR where
+    hashWithSalt salt (FileR f _) = hashWithSalt salt f
+
 instance NFData FileA where
     rnf (FileA a b c) = rnf a `seq` rnf b `seq` rnf c
+
+instance NFData FileR where
+    rnf (FileR f _) = rnf f
 
 instance Binary FileA where
     put (FileA a b c) = put a >> put b >> put c
     get = liftA3 FileA get get get
 
+instance Binary FileR where
+    put (FileR f _) = put f
+    get = (`FileR` False) <$> get
+
 instance Show FileA where
     show (FileA m s h) = "File {mod=" ++ show m ++ ",size=" ++ show s ++ ",digest=" ++ show h ++ "}"
+
+instance Show FileR where
+    show (FileR f b) = show f ++ if b then " recomputed" else " not recomputed"
 
 instance Storable FileA where
     sizeOf _ = 4 * 3 -- 4 Word32's
@@ -182,16 +201,16 @@ defaultRuleFile = do
     -- A rule from FileQ to (Maybe FileA). The result value is only useful for linting.
     addBuiltinRuleEx (ruleLint opts) (ruleRun opts $ shakeRebuildApply opts)
 
-ruleLint :: ShakeOptions -> BuiltinLint FileQ (Maybe FileA) 
-ruleLint opts k Nothing = return Nothing
-ruleLint opts k (Just v) = do
+ruleLint :: ShakeOptions -> BuiltinLint FileQ FileR
+ruleLint opts k (FileR Nothing _) = return Nothing
+ruleLint opts k (FileR (Just v) _) = do
     now <- fileStoredValue opts k
     return $ case now of
         Nothing -> Just "<missing>"
         Just now | fileEqualValue opts v now == EqualCheap -> Nothing
                  | otherwise -> Just $ show now
 
-ruleRun :: ShakeOptions -> (FilePath -> Rebuild) -> BuiltinRun FileQ (Maybe FileA)
+ruleRun :: ShakeOptions -> (FilePath -> Rebuild) -> BuiltinRun FileQ FileR
 ruleRun opts@ShakeOptions{..} rebuildFlags o@(FileQ x) oldBin@(fmap getEx -> old) dirtyChildren = do
     -- for One, rebuild makes perfect sense
     -- for Forward, we expect the child will have already rebuilt - Rebuild just lets us deal with code changes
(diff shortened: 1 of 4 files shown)


erwo42:
This should enable some use cases from #502

erwo42:
Seems like needHasChanged as implemented here is broken for targets produced by rules created by fileForward, as the rules created by fileForward don't forward the information, wether the target was rebuild. Will try to fix this.

erwo42:
This fixed the needHasChanged for rules created by fileForward. So you can use needHasChanged for a target produced by a `&%>` or `&?>` rule.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -58,6 +58,13 @@ newtype FileQ = FileQ {fromFileQ :: FileName}
 data FileA = FileA {-# UNPACK #-} !ModTime {-# UNPACK #-} !FileSize FileHash
     deriving (Typeable,Eq)
 
+-- | Result of a File rule, may contain raw file information and wether the rule did run this build
+data FileR = FileR (Maybe FileA) Bool
```
ndmitchell: I'd have probably gone for a pair, although in many ways a data type is "cleaner". If going for a data type, I'd use record names, so `hasChanged` becomes a record, and also an annotation on the data type.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -87,16 +94,29 @@ instance BinaryEx [FileQ] where
...
     hashWithSalt salt (FileA a b c) = hashWithSalt salt a `xor` hashWithSalt salt b `xor` hashWithSalt salt c
 
+instance Hashable FileR where
+    hashWithSalt salt (FileR f _) = hashWithSalt salt f
+
 instance NFData FileA where
     rnf (FileA a b c) = rnf a `seq` rnf b `seq` rnf c
 
+instance NFData FileR where
+    rnf (FileR f _) = rnf f
```
ndmitchell: NFData should definitely evaluate both pieces - the purpose of NFData is to flush out all exceptions, so everything needs covering.

src/Development/Shake/Internal/Rules/File.hs
ndmitchell: Also make sure to say the Bool isn't serialised and gets recreated at False, doesn't participate in Eq/Hash etc.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -313,6 +333,32 @@ fileForward act = addUserRule $ FileRule $ fmap ModeForward . act
...
+--
+--    It can be used together with 'Development.Shake.&%>' or 'Development.Shake.&?>'
+--    to implement batching. There is at
+--    the moment no way to get a list of inconsistent targets in such a rule, therefore
+--    either all targets need to be rebuild or it needs to be ensured, that targets don't
+--    get inconsistent. (This happens if targets get altered in a build after they have
+--    been produced or if some process alter the targets between builds. Support for
+--    detecting inconsistent targets will be added later.)
+needHasChanged :: [FilePath] -> Action [Bool]
+needHasChanged = fmap (fmap hasChanged) . apply_ fileNameFromString
```
ndmitchell: As agreed, returning the `[FilePath]` values that changed is a clearer API.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -58,6 +58,13 @@ newtype FileQ = FileQ {fromFileQ :: FileName}
 data FileA = FileA {-# UNPACK #-} !ModTime {-# UNPACK #-} !FileSize FileHash
     deriving (Typeable,Eq)
 
+-- | Result of a File rule, may contain raw file information and wether the rule did run this build
+data FileR = FileR (Maybe FileA) Bool
+    deriving (Typeable, Eq)
```
ndmitchell: Can you derive `Eq`? Don't you need to change it to ignore the Bool? In fact, I'm slightly surprised that it doesn't break at least one test if you look at both things...

erwo42: I didn't understand the whole code base, but I would conjecture rule results aren't used as hashmap keys and aren't usually compared for equality? What would one look up by a rule result? And the comparison in the BuiltinRule is very specialised.
Maybe that's why it didn't matter in the tests?

src/Development/Shake/Internal/Rules/File.hs
erwo42: Of course. My bad!

src/Development/Shake/Internal/Rules/File.hs
erwo42: I chose the record for now. I can change that to a pair, if you prefer that.

erwo42: I remember now, why I chose a plain data type last time. Travis slaps me for not using the first record field ...

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -58,6 +58,12 @@ newtype FileQ = FileQ {fromFileQ :: FileName}
 data FileA = FileA {-# UNPACK #-} !ModTime {-# UNPACK #-} !FileSize FileHash
     deriving (Typeable,Eq)
 
+-- | Result of a File rule, may contain raw file information and wether the rule did run this build
+data FileR = FileR { result :: Maybe FileA -- ^ Raw information about the file built by this rule. Set to 'Nothing' to prevent linting some times.
+                   , hasChanged :: Bool    -- ^ Wether the file changed this build. Transient information, that doesn't get serialized, is recreated at 'False' and doesn't participate in 'Eq', 'Hash' and the like.
```
ndmitchell: Spelling mistake - `Whether`. Probably also worth wrapping these lines at some point (I'm not a rigid follower of any particular length - but these lines are pretty long)

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -87,16 +93,32 @@ instance BinaryEx [FileQ] where
 instance Hashable FileA where
     hashWithSalt salt (FileA a b c) = hashWithSalt salt a `xor` hashWithSalt salt b `xor` hashWithSalt salt c
 
+instance Eq FileR where
+   fr1 == fr2 = result fr1 == result fr2
+
+instance Hashable FileR where
+    hashWithSalt salt (FileR f _) = hashWithSalt salt f
```
ndmitchell: The `Hashable` and `Eq` should be consistent - which they are. But one uses pattern matching and one uses `result` selector - they should probably both be the same style.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -313,6 +335,30 @@ fileForward act = addUserRule $ FileRule $ fmap ModeForward . act
...
+--   to be rebuild.
+--
+--   Notice that a rule can be run even if no dependency has changed and needHasChanged
+--   will return an empty list on every call then. In this case a target got inconsistent
+--   and you should recreate it, without reusing any part of the target from the last build,
+--   as the target could have been edited or even deleted.
+needHasChanged :: [FilePath] -> Action [FilePath]
+needHasChanged paths = do
+    changed <- fmap hasChanged <$> apply_ fileNameFromString paths
+    return $ fmap fst $ filter snd $ zip paths changed
```
ndmitchell: Both those `fmap`'s are secret `map`s, so better to be explicit. However, I'd probably do:

```
res <- apply_ fileNameFromString paths
return [a | (a,b) <- zip paths res, hasChanged b]
```

That seems cleaner.

erwo42:
Please wait with merging. I will fix those issues and add some test cases.

src/Development/Shake/Internal/Rules/File.hs
erwo42: That is a lot clearer.

erwo42:
I can't reproduce these failures at my local machine (linux) and the travis build seems a bit erratic, so I suspect that is due some specialties about windows file timestamps, which I don't know about.
Should I just add "--sleep" to each "build" line in the tests sheepishly or do you have a suspicion what went wrong?

ndmitchell:
I'd suggest putting all the tests in `Test.Batch` as these are really the batch feature. For the `--sleep`, generally if you modify a twice (e.g. `writeFile`), they must be separated by a build that does `--sleep`. The reason is that some filesystems have accuracies as low as 1 second, and `--sleep` sleeps for the minimum sleep change.

ndmitchell:
As a heuristic, just add `--sleep` everywhere to start with, and remove it later. On modern OS's it adds a few milliseconds delay at most.

ndmitchell:
Reading through, the Eq/Hashable/Binary classes aren't used, so it's pretty nasty to have to implement them, and nearly impossible to come up with a "good" semantics for them. I've just pushed some refactorings that let them become dead code, so after your patch is merged, I'll go through and delete those classes for FileA/FileR etc.

erwo42:
Regarding the tests I would not concur. Sorry for the following wall of text.

The tests in Test.Basic test whether `needHasChanged` detects changed dependencies and whether it notices the right one.

The last test in Test.Basic demonstrates, that there is an open problem regarding the use case in #502 (`slowCompile`, `fastUpdate`) under the assumption, that you don't want to loose shakes nice property of detecting targets, with which has been messed. If no dependency is changed and the `Action()` is run, you can assume the target was messed with and rebuild it by slowCompile (as it can be altered in unanticipated ways). But if one changes the dependencies in a way, that triggers `fastUpdate` and the target was messed with, your screwed. At least I don't see, how you could detect this at the moment. (But I still think, my PR is a useful first step.)

The tests in Test.Files mix several things (and I openly admit, that one of those things could be considered a "static batch").

First, we want to see, that the rules created by `Development.Shake.Internal.Rules.File.fileForward` transport the information about rebuild targets (as dependencies of other rules) properly and fine grained. (As opposed to a situation, were all targets of a multiple target rule are always considered rebuild or not together.)

Secondly we demonstrate the same problem as in `Test.Basic`. Of course in this case, that looks like some kind of "static batch", but If the problem was solved I might imagine that `needHasChanged` could have similar uses like the use case "slowCompile, fastUpdate" for single target rules.

Should I restate some of these things in #502 ?

I also would like to make a case, that a "static batch" might be a useful alternative (although in less cases) to `batch` (which I think is a great thing). But I would open another ticket for that at a later time (need sleep now :) ).

Thanks for your time and considering my stuff here!

ndmitchell:
For info, before seeing you reasons, I moved the basic tests into batch. Suggest we figure out an answer, then I'll move them back, or to a new location.

I think my reasoning is quite different. From a user perspective, both `needHasChanged` and `batch` let users take individual work and slice it up differently - reasonably similar operations. A user who comes wanting `needHasChanged` might well discover they secretly want `batch` and vice versa - so they are both the "batching" features of Shake - even if internally they are really different.

Given the tests in the two modules are basically the same, they seem like they should live together? Which gives the idea that either they should both go in Batch, or both go somewhere new holding them both.

Regarding the bug you outline, agreed, it's a tricky case - sometimes you want to rebuild not because the dependencies have changed, but because the target has. Fortunately the people who are messing around with `needHasChanged` are probably not standard users, so a bit of a rough edge isn't the end of the world as a first step.

erwo42:
I'll see your point. Let's shuffle the tests to `Test.Batch`.

PR #541, opened 2017-10-16
fixed a missing import and probably more stuff, until travis is happy again

erwo42:


erwo42:
Hmm 0.3 s faster ninja build when shake is compiled with ghc 7.4.2, don't know how to fix that.

ndmitchell:
Occasionally the Ninja build is faster because the VM is somewhat unpredictable. In those cases I tend to just rerun it - it isn't anything fundamental - just that performance testing is hard to do.

Thanks for the fix!

PR #542, opened 2017-10-16
fixed misfiled tests

erwo42:
some code duplication though, maybe open up `Test.Files.Internal` and put `&?%>` and `Args` there?


ndmitchell:
There is Test.Type which usually serves for the repository of all helper functions for the test suite - those definition could certainly live there.

PR #544, opened 2017-10-21
fixed typo in comment

erwo42:
this made copy and pasting awful


ndmitchell:
Thanks! As it happened, I had made this an explicitly whitelisted comment in the doc test, so I've updated that too (it explains the test failure).

Issue #550, opened 2017-11-13
Extract FilePattern to its own library


eborden:
Shake's `FilePattern` seems like a useful module. It should be extract into a library so it can be shared by other tools.

This has been mentioned before:
https://github.com/ndmitchell/shake/issues/499
https://github.com/ndmitchell/shake/issues/485
https://github.com/ndmitchell/hlint/issues/402
https://github.com/ndmitchell/hlint/pull/334

ndmitchell:
Are you going to give this a try?

ndmitchell:
One thing to note is that Shake supports `//` patterns mostly as a legacy feature. I suggest the new `filepattern` package also supports that, but perhaps behind a `useLegacyPatterns` function, just so Shake can migrate more slowly.

eborden:
I'm going to give it a go.

eborden:
First pass to extract the necessary modules:
https://github.com/eborden/filepattern

eborden:
I've done the work to provide `FilePattern` and `FilePattern.Legacy`. As well as opening #551 to begin relying on that library. There is a bit more work in `filepattern` getting CI running and re-enabling some tests, but should be enough draft work for you to review.

eborden:
Oh also not sure how you want to handle the `filepattern` repo. Totally fine to hand it over to you if you'd rather keep maintainer-ship.

ndmitchell:
Thanks for all the work, I’ll review later. What do you prefer/hope in terms of maintainership? You, me or both seems the possible options.

eborden:
I think a dual maintainership sounds good. This lib is integral to `shake` so it is probably good for you to at least have one hand on the wheel.

eborden:
Popping in here to see if you've had a chance to review.

ndmitchell:
Alas, not yet - I've penciled in a train journey tomorrow morning to do a thorough review!

eborden:
@ndmitchell How are you feeling about this initiative?

ndmitchell:
Still positive, going through it slowly. Unfortunately other things cropped up so it's currently 5th on my 3rd todo list (I work through them all in parallel).

ndmitchell:
Completed - thanks to your help and constant enthusiasm. https://hackage.haskell.org/package/filepattern is the result, and it's vastly better than the stuff in Shake.

eborden:
Awesome!

ndmitchell:
[...]

PR #551, opened 2017-11-14
Ulitize FilePattern library

eborden:
This addresses https://github.com/ndmitchell/shake/issues/550

FilePattern has been extracted to a library. Its functionality was
useful elsewhere. The library itself is dropping support for "//"
patterns, but they remain available in the `FilePattern.Legacy` module
so shake can have a slower path away from them.

The current draft of `filepattern` is here:
https://github.com/eborden/filepattern

This PR will not build until that library has been released on hackage.


ndmitchell:
The first thing to do is thoroughly review the FilePattern library, which I've started doing, and have pushed a bunch of commits to https://github.com/ndmitchell/filepattern. Some of those will necessarily result in tweaks to this patch, but I'd wait until the FilePattern stuff is done first (I expect to make more tweaks this morning). There are still a few tweaks I want to make, but it looks really good. Two questions:

* Should it be `Text.FilePattern`? or `System.FilePattern`?
* I prefer to avoid "internalTest" things inside the main module, so the options are have a `Text.FilePattern.Internal` for reexporting just the test-specific bits, try to fudge it on the outside, or just compile the code multiple times - so the test doesn't depend on the source (cleanest, and usually my preference). Do you have any thoughts?

eborden:
@ndmitchell 
1. Yeah, I wasn't sure what would be best here. It is dependent on `env` so `System` might be a better location.
2. Yup, that makes sense to me.

ndmitchell:
To update where things are, as I wander through filepattern I'm making some improvements to it, and also taking the change to rethink certain aspects (mostly minor, but things that I don't want to put in stone while wrong). I'm hopefully under a week away from being happy with the filepattern library, and then actually flipping Shake to it shouldn't be too hard.

ndmitchell:
I've now rewritten a lot of the `filepattern` library, so the interfaces and semantics are no longer what they were when the library left Shake. Given that the stringy things have changed in ways that will break code, but not at compile time, and that `//` patterns got deleted entirely, I'm wondering how to make the switch. My inclination is to get `filepattern` into HLint first, confirm it doesn't do anything super stupid, and then switch in Shake, with a very deliberate 0.1 bump and lots of publicity.

eborden:
@ndmitchell that sounds like a reasonable plan. Bugs can be sussed out as users discover them in `hlint` instead of a big bang in `shake`.

ndmitchell:
Closing this PR as given the extensive changes in `filepattern`, I don't think it's going to provide much useful benefit in the migration. The ticket #550 should definitely remain for now, until we're ready to go forward.

PR #553, opened 2017-11-21
Don't store new FileQs if nothing changed

duog:
Storing the new FileQ was forcing all of the hash thunks.

I'm quite unfamiliar with shake, so apologies if I'm mistaken. I saw that all files were being hashed while profiling hadrian. I verified by `putStrLn`-instrumenting the `unsafeInterleaveIO` thunk in `fileStoreValue`. This speeds up a no-op hadrian run by about 15% (-j1, and profiled).

Sorry there's no test, I wasn't up to working out how the testing framework works.

### DIFF
--- a/src/Development/Shake/Internal/Rules/File.hs
+++ b/src/Development/Shake/Internal/Rules/File.hs
@@ -234,7 +234,7 @@ ruleRun opts@ShakeOptions{..} rebuildFlags o@(FileQ x) oldBin@(fmap getEx -> old
             case now of
                 Nothing -> rebuild
                 Just now -> case fileEqualValue opts old now of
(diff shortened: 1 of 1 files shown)


ndmitchell:
Wow, impressive catch! And almost impossible to write a test for - but I'll certainly try given how fundamental it is.

duog:
You can set file modification time (on at least some platforms) with `System.Directory.setModificationTime`, I think one should be able to trigger it with that.

ndmitchell:
Sorry it took so long, but this is a good patch, and definitely improves things - nice work!

While I can trigger the behaviour, I can't write a test asking whether the file gets hashed or not, since it's hidden in the depths of Shake and is (entirely deliberately!) unobservable. Never mind, a great change.

PR #563, opened 2018-02-02
Improve docs for getEnvWithDefault

markus1189:
Found the old style of documenting the arguments a little cumbersome, so I took the liberty to change it to a style I found in other packages.


ndmitchell:
Thanks! Agreed - much clearer.

PR #584, opened 2018-04-19
Add semigroup instance for Action

vmchale:
This would add `Semigroup` and `Monoid` instances for `Action`, keeping compatibility with past compiler versions.


ndmitchell:
Thanks - a good idea! Shake already depends on the `semigroups` library on old versions of GHC, so I was able to delete a lot of the CPP.

PR #590, opened 2018-05-01
Print "Build completed" message with millisecond accuracy

bitc:
This changes the "Build completed" message.

If the time is at least one minute then the output remains the same:

```
Build completed in 2:34m
```

But if the time is less than one minute then the output has millisecond accuracy:

```
Build completed in 12.624s
```

```
Build completed in 0.057s
```

This added accuracy is useful when you are optimizing the zero build of your build system

### DIFF
--- a/src/Development/Shake/Internal/Args.hs
+++ b/src/Development/Shake/Internal/Args.hs
@@ -204,8 +204,11 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
                 Right () -> do
                     tot <- start
                     let (mins,secs) = divMod (ceiling tot) (60 :: Int)
-                        time = show mins ++ ":" ++ ['0' | secs < 10] ++ show secs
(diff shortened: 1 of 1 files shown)


ndmitchell:
I did that for compatibility with `make`, but agreed that changing it is a good idea. I'd be tempted to use [`showDuration`](https://hackage.haskell.org/package/extra-1.6.6/docs/Extra.html#v:showDuration) which gives you precision to 1/100th of a second, is already written/tested, and is used elsewhere in Shake.

For debugging the zero time I suggest using `--timing` - that already gives you millisecond precision, and also records where the time went so you know if you hit a database compaction taking an addition 0.004` seconds. See https://shakebuild.com/profiling#timing-information.

ndmitchell:
I followed what I suggested and modified your patch to use `showDuration` after.

PR #593, opened 2018-05-14
Correct a github URL.

EdwardBetts:


ndmitchell:
Thanks! I had actually intended it to link all the way to https://github.com/ndmitchell/shake/releases/tag/shake-progress-1, but clearly wasn't having a good day when I wrote that URL...

PR #599, opened 2018-05-19
Test 8.4.2 in travis-ci

mitchellwrosen:


ndmitchell:
Cool! Any reason to test with 8.4.2 and 8.4.1? I usually retire old point releases to be friendly to Travis.

All the CI tests fail because I have a central list of what versions I test with for all my packages. I'll update that tonight and cascade to my packages.

ndmitchell:
I pushed a patch that just modified 8.4.1 to 8.4.2. Thanks for pointing it out!

mitchellwrosen:
Of course!

PR #618, opened 2018-09-29
Adds known to allow checking for unknown targets

typetetris:
Using known one can implement a check for files, which can be build by
a rule database. With this information one can delete files, not buildable
by a shake build system. Hopefully those are stale artifacts

To elaborate: Under a bunch of assumptions, one can implement a rule to delete stale targets.
For example: Consider a build system, which starts with an empty build directory and each file, which is created there, is created by a rule within the shake build system. In this case one can check every file in the build directory with known and delete those, that aren't known any more.

An approximation could have been implemented by needing all the files in the build directory again, but then they also would be rebuild, if they are out of date and the build would stop for each unknown file und you would have to reiterate.

Maybe if known is a to great risk to add to the official api, one could place it in a Development.Shake.Experimental module and state, that such a module contains parts of the api with a high probability of changing or evening disappearing.

Thanks for the pull request!

### DIFF
--- a/src/Development/Shake.hs
+++ b/src/Development/Shake.hs
@@ -103,6 +103,7 @@ module Development.Shake(
     needHasChanged,
     resultHasChanged,
     batch,
+    known,
     -- * Deprecated
     (*>), (|*>), (&*>),
     (**>), (*>>), (?>>),
--- a/src/Development/Shake/Internal/Rules/File.hs
+++ b/src/Development/Shake/Internal/Rules/File.hs
@@ -6,7 +6,7 @@ module Development.Shake.Internal.Rules.File(
     trackRead, trackWrite, trackAllow,
     defaultRuleFile,
     (%>), (|%>), (?>), phony, (~>), phonys,
-    resultHasChanged,
+    resultHasChanged, known,
     -- * Internal only
     FileQ(..), FileA(..), fileStoredValue, fileEqualValue, EqualCost(..), fileForward
     ) where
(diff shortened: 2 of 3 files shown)


typetetris:
The AppVeyor failure seems unrelated with my changes.

ndmitchell:
Thanks for the patch! The Appveyor failure is indeed not your fault. The code looks fine, and thanks for the test, so just a few questions on the motivation. 

What if you have a rule for `*.txt` ? Then it will note that all txt files are valid, even though some may have had their dependencies deleted.  Have you seen http://neilmitchell.blogspot.com/2015/04/cleaning-stale-files-with-shake.html?

typetetris:
If I have a rule for `*.txt` all txt files will indeed be considered valid.

Using `known` for removing stale targets needs, as stated above, a bunch of assumptions or guarantees from the way you model your build system, to be useful. But if you can use it and its approximation is good enough, you don't need to update all targets (as I understood you need to do with `--live`), just to find the stale ones.

Of course my first argument of reiteration was bogus, as `--live` deals with that perfectly.

Imagine a build system with many targets and each individual developer needs to build just a few of them in their day to day workflow. You want to be able to remove stale targets in that subgraph of dependencies, the individual developers needs, without deleting all the targets, he only needs to build occasionally (like every two weeks or so).

I will give it some more thought and close the pull request for now. Maybe I reopen it later, with better arguments or more experience.

PR #625, opened 2018-11-05
Add: getTargets :: ShakeOptions -> Rules () -> IO [String]

basvandijk:
Hi Neil, as promised I added:

```haskell
getTargets :: ShakeOptions -> Rules () -> IO [String]
```

`getTargets` returns the targets that are registered by `phony`, `~>`,   `%>`, `|%>`  and `&%>`.

As discussed, this is useful for implementing shell autocompletion of targets for example.

Future work
-----------------
A future extension could be to add a combinator that can associate some documentation with a Rule. Then we can imagine a function:

```haskell
getTargetsWithDocs :: ShakeOptions -> Rules () -> IO [(String, Maybe String)]
``` 

which is like `getTargets` but pairs the targets with some optional documentation. This documentation can then be used in shell autocompletion for example for shells that support that.

The combinator to add documentation could look like:

```haskell
doc :: String -> Rules () -> Rules ()
```

The semantics is that `doc txt r` will associate `txt` with every target in `r`. Inner calls of `doc` will overwrite outer calls. So `doc txt1 (doc txt2 r)` will associate `txt2` with every target in `r`.

I'm not sure this is the right API yet. It needs more thought and experimentation.

### DIFF
--- a/src/Test/GetTargets.hs
+++ b/src/Test/GetTargets.hs
@@ -1,0 +1,63 @@
+module Test.GetTargets(main) where
+
+import Development.Shake
+import Test.Type
+
+main :: IO () -> IO ()
+main _sleeper = do
+    targets <- getTargets shakeOptions rules
+    targets === expected
+
+rules :: Rules ()
+rules = do
+  phony "phony1" $ return ()
+
+  "file1" %> \_ -> return ()
+  ["file2", "file3"] |%> \_ -> return ()
+  ["file4", "file5"] &%> \_ -> return ()
+
+  build "file6" $ \_ -> return ()
+  buildAny ["file7", "file8"] $ \_ -> return ()
+  buildAll ["file9", "file10"] $ \_ -> return ()
+
+  phonyWithDocs
+    "Builds something really good"
+    "phony2" $ return ()
+
+  buildWithDocs
+    "a great file"
+    "file11" $ \_ -> return ()
+
+  buildAnyWithDocs
+    "awesome files"
+    ["file12", "file13"] $ \_ -> return ()
+
+  buildAllWithDocs
+    "just briliant"
+    ["file14", "file15"] $ \_ -> return ()
+
+expected :: [Target]
+expected =
+    [ Target "phony1" (Just "A phony target")
+
+    , Target "file1"  Nothing
+    , Target "file2"  Nothing
+    , Target "file3"  Nothing
+    , Target "file4"  Nothing
+    , Target "file5"  Nothing
+
+    , Target "file6"  Nothing
+    , Target "file7"  Nothing
+    , Target "file8"  Nothing
+    , Target "file9"  Nothing
+    , Target "file10" Nothing
+
+    , Target "phony2" (Just "Builds something really good")
+
+    , Target "file11" (Just "a great file")
+    , Target "file12" (Just "awesome files")
+    , Target "file13" (Just "awesome files")
+    , Target "file14" (Just "just briliant")
+    , Target "file15" (Just "just briliant")
+
+    ]
--- a/shake.cabal
+++ b/shake.cabal
@@ -448,6 +448,7 @@ test-suite shake-test
         Test.FilePattern
         Test.Files
         Test.Forward
+        Test.GetTargets
         Test.History
         Test.Journal
         Test.Lint
--- a/src/Development/Ninja/All.hs
+++ b/src/Development/Ninja/All.hs
@@ -5,7 +5,7 @@ module Development.Ninja.All(runNinja) where
 import Development.Ninja.Env
 import Development.Ninja.Type
 import Development.Ninja.Parse
-import Development.Shake hiding (addEnv)
+import Development.Shake hiding (addEnv,build)
 import Development.Shake.Classes
 import qualified Data.ByteString as BS8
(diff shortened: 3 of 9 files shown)


src/Test/GetTargets.hs
```diff
@@ -0,0 +1,28 @@
+module Test.GetTargets(main) where
+
+import Control.Monad
+import Development.Shake
+import Test.Type
+
+main :: IO () -> IO ()
+main _sleeper = do
+  targets <- getTargets shakeOptions rules
+  unless (targets == expected) $
```
ndmitchell: I have `===` in Test.Type which abstracts this and gives nice errors etc. Easier to use that.

ndmitchell:
Thanks for the PR - nice work!

My initial remark would be that I wasn't expecting patterns to go in the targets list - but thinking further you can remove them by filtering and testing for patterns (a function I'm going to add and expose through another library), so better to have them.

I think the idea of docs is an important one, and I suggest we put it in now, so we don't have to change the API later. My first though in this direction is:

* Make the data type `[(String, Maybe String)]` to capture the docs.
* Make `addTarget :: String -> Maybe String -> Rules ()`, and in all places we call it automatically use `Nothing`.
* Add `withoutTargets :: Rules a -> Rules a` which removes all targets in a block of rules, much like `withoutActions`.

With that, if you do want documentation, you can do:

```haskell
addTarget "named" "my phony with docs"
withoutTargets $ phony "named" ...
```

We might want to add a special combinator `phonyWithDocs` (or a better name if we can think of it) since I imagine phony's will be the most common thing to add docs for.

We might even want to default the docs on phony to be "A phony target"? Not sure there - but there seems some value in showing things are phony vs real by default, since they behave very differently.

src/Test/GetTargets.hs
```diff
@@ -0,0 +1,28 @@
+module Test.GetTargets(main) where
+
+import Control.Monad
+import Development.Shake
+import Test.Type
```
ndmitchell: FWIW this import is currently redundant, and is why all the tests fail. However, with my suggestion below it would be used once more.

basvandijk:
Thanks for the review!

> My initial remark would be that I wasn't expecting patterns to go in the targets list 

For discoverability and comfort it's quite useful if paths are autocompleted. In our `lumi` tool it is the case that patterns are completed literally. So for example we have the following completion:

```
$ lumi os/users/\*/password.enc 
```

Note that currently the pattern gets completed literally, i.e. including the `*`. In this case the user has to move back his cursor and substitute the `\*` for the right value. This could be considered a bit irritating but on the other hand the user immediately sees the whole target so he doesn't have to remember it nor type it in full. I can also imagine support for smarter autocompletion being build on top of this.

> but thinking further you can remove them by filtering and testing for patterns (a function I'm going to add and expose through another library), so better to have them.

That would be a nice feature.

> I think the idea of docs is an important one, and I suggest we put it in now, so we don't have to change the API later. My first though in this direction is:
>
> * Make the data type [(String, Maybe String)] to capture the docs.

Agreed.

> * Make addTarget :: String -> Maybe String -> Rules (), and in all places we call it automatically use Nothing.

Agreed.

>* Add withoutTargets :: Rules a -> Rules a which removes all targets in a block of rules, much like withoutActions.
>
>With that, if you do want documentation, you can do:
>```
>addTarget "named" "my phony with docs"
>withoutTargets $ phony "named" ...
>```

In the latter example we have our original problem again where targets are repeated for both the rule and the documentation. This can get out of sync quickly and thus prone to errors. I think the solution is in your next suggestion:

>We might want to add a special combinator `phonyWithDocs` (or a better name if we can think of it) since I imagine phony's will be the most common thing to add docs for.

Yes, I like that.

> We might even want to default the docs on phony to be "A phony target"? Not sure there - but there seems some value in showing things are phony vs real by default, since they behave very differently.

I think that makes sense as well.

So what about not exposing `addTarget` like we have now and instead of providing `withoutTargets` we provide `phonyWithDocs`? 

The last remaining question is what to do about the file-pattern rules: `%>`, `|%>` and `&%>`? Add `*WithDocs` alternatives for these as well?

We might first consider adding function synonyms for these operators. Then we can add the `*WithDocs` alternatives after that. For example:

```
match    = (%>)
matchAny = (|%>)
matchAll = (&%>)
```

Or I think I like these more:

```
build    = (%>)
buildAny = (|%>)
buildAll = (&%>)
```

ndmitchell:
Not a fan of the `buildAnyWithDocs` approach - it means that every Shake function now comes in 3 flavours - an operator, a named and a namedWithDocs. For `phony` I think that's fine, it's by an overwhelming majority the most common. However, I wonder if:

```
withDocs :: String -> Rules a -> Rules a
```

Might be easier? `withDocs "my docs" $ phony "foo" $ ...` reads reasonably well, and is compositional, requiring no additional values. I think still having `addTarget` and `withoutTargets` is valuable, as those are the pieces that allow people to write their own combinators if they want them.

ndmitchell:
Thanks! I switch to using fewer combinators, so the end set of new functions are:

> getTargets, addTarget, withTargetDocs, withoutTargets,

That seems a lot cleaner, and isn't much difference when expressing the tests, so I'm happy with it. Thanks for the PR!

PR #630, opened 2018-12-01
Fix comment

snowleopard:
Thanks for the pull request!

### DIFF
--- a/src/Development/Shake/Internal/Core/Action.hs
+++ b/src/Development/Shake/Internal/Core/Action.hs
@@ -354,7 +354,7 @@ historyDisable :: Action ()
 historyDisable = Action $ modifyRW $ \s -> s{localHistory = False}
 
 
--- | This rule the following files, in addition to any defined by its target.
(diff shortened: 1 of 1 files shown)


snowleopard:
@ndmitchell Or "produces" :)

ndmitchell:
@snowleopard Thanks for the patch, builds is good, since it helps in addition to the name of the function.

There is no tracking that the `produces` things exist when doing a rebuild check - there probably should be. I created https://github.com/ndmitchell/shake/issues/631 to track that. So far `produces` is really intended for the cloud build - it copies the files to the server if you have called `produces`.

However, I don't see how this is related to the trac ticket?

snowleopard:
> However, I don't see how this is related to the trac ticket?

In the Trac ticket we have a situation when we don't know build results of a rule upfront: we unpack an archive whose contents is unknown and it would be nice to be able to dynamically record this using a tracked version of `produces`.

(But I don't know whether this is actually related to the lint failure.)

snowleopard:
Great! I've added a comment to #631 to explain the use-case.

PR #648, opened 2019-02-13
Depend on heaps >= 0.3.6.1

RyanGlScott:
The `heaps` was dependency was previously removed due to concerns over its `cabal-doctest` dependency (ekmett/heaps#13). That dependency was removed in https://github.com/ekmett/heaps/commit/571329f710addee1b6ef1bd5a7116755bf20af30 and released as `heaps-0.3.6.1`, so `shake` should be able to depend on it once more.


ndmitchell:
Thanks!

PR #654, opened 2019-02-27
WIP Democratize shared cache implementations

pepeiborra:
This change defines two datatypes `SharedCache` (internal) and `SharedFileSystem` (api) that allow users of Shake to define their own stores of cache history data.

This is only a draft, as I was unsure of whether it would be accepted and didn't want to invest too much time into it. Also the manual `Data` instances really threw me off :)

### DIFF
--- a/src/Development/Shake/Internal/History/Shared.hs
+++ b/src/Development/Shake/Internal/History/Shared.hs
@@ -1,14 +1,16 @@
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE RecordWildCards, TupleSections #-}
 
 module Development.Shake.Internal.History.Shared(
-    Shared, newShared,
-    addShared, lookupShared,
-    removeShared, listShared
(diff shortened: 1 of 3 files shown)


ndmitchell:
Thanks for the patch!

Yeah, I rather wish I hadn't started down the manual data instances route, but now I've done that route, it's hard to turn back...

I think the basic idea of wanting an http and a file endpoint, and thus opening it up makes sense. However, ShakeOptions deliberately tries to mirror command line arguments where possible. I wonder if anyone will ever need something that isn't http or filesystem? If so, I'd rather make shakeShare (or shakeCloud?) be a list of strings, and treat http:// ones specially - and then do the abstraction layer you've done entirely inside.

pepeiborra:
Will anyone need something other than filesystem or http ? Probably no. One benefit of this change is that it takes you off the hook for implementing http, but it sounds like you are already on it.

I'll close the PR then unless you want to do something else with it

PR #656, opened 2019-03-07
Calculate correct relative paths from fsatrace output.

DavidEichmann:
I've adding this PR in relation to GHC trac [issue 16400](https://ghc.haskell.org/trac/ghc/ticket/16400).

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -145,7 +145,12 @@ commandExplicit funcName oopts results exe args = do
         shelled = runShell (unwords $ exe : args)
 
         ignore = map (?==) shakeLintIgnore
-        ham cwd xs = [makeRelative cwd x | x <- map toStandard xs
+
+        -- makeRelativeEx is used instead of makeRelative. Only makeRelativeEx
+        -- is guaranteed to give a valid relative path as long as the input
+        -- paths are on the same drive.
+        ham cwd xs = liftIO $ sequence [fromMaybe x <$> makeRelativeEx cwd x
+                                         | x <- map toStandard xs
                                          , any (`isPrefixOf` x) shakeLintInside
                                          , not $ any ($ x) ignore]
 
@@ -189,8 +194,8 @@ commandExplicit funcName oopts results exe args = do
                 existing f = liftIO . filterM doesFileExist . nubOrd . mapMaybe f
             rs <- existing reader xs
             ws <- existing writer xs
-            let reads = ham cwd rs
-                writes = ham cwd ws
+            reads  <- ham cwd rs
+            writes <- ham cwd ws
             when useAutoDeps $
                 unsafeAllowApply $ needed reads
             trackRead reads
@@ -202,7 +207,7 @@ commandExplicit funcName oopts results exe args = do
             pxs <- liftIO $ parseFSAT <$> readFileUTF8' file
             xs <- liftIO $ filterM doesFileExist [x | FSATRead x <- pxs]
             cwd <- liftIO getCurrentDirectory
-            unsafeAllowApply $ need $ ham cwd xs
+            unsafeAllowApply $ need =<< ham cwd xs
             return res
 
     skipper $ tracker $ \exe args -> verboser $ tracer $ commandExplicitIO funcName opts results exe args
--- a/src/Development/Shake/FilePath.hs
+++ b/src/Development/Shake/FilePath.hs
@@ -7,11 +7,12 @@
 module Development.Shake.FilePath(
     module System.FilePath, module System.FilePath.Posix,
     dropDirectory1, takeDirectory1, replaceDirectory1,
-    normaliseEx,
+    makeRelativeEx, normaliseEx,
     toNative, toStandard,
     exe
     ) where
 
+import System.Directory (canonicalizePath)
 import System.Info.Extra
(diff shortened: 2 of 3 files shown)


src/Development/Shake/Command.hs
```diff
@@ -145,7 +145,8 @@ commandExplicit funcName oopts results exe args = do
         shelled = runShell (unwords $ exe : args)
 
         ignore = map (?==) shakeLintIgnore
-        ham cwd xs = [makeRelative cwd x | x <- map toStandard xs
+        ham cwd xs = liftIO $ sequence [fromMaybe x <$> makeRelativeEx cwd x
```
DavidEichmann: I'm concerned with performance here. Thoughts?

DavidEichmann: What does `ham` actually mean?
There is an underlying assumption here that track uses relative paths as keys. Is that correct? It's troubling to me that there may be many ways to refer to the same file (relative, absolute, via different symlinks or indirections), yet there doesn't seem to be any mechanism to enforce a canonical usage.

ndmitchell: We only call this after spawning a process. Hard to do much that has a performance impact relative to that.

ndmitchell:
Thanks for the patch. Can you explain the output you get before this patch? e.g. when running you see XXX, and we're relative to YYY, and you hoped to make it ZZZ but it failed? We should generally in Shake be referring to things by relative paths. If you have symlinks, you are going to have problems with build systems. If you use absolute paths its much harder to be portable for features like cloud shake. Relative paths everywhere solves these problems nicely.

DavidEichmann:
I've added a bit of documentation to the commit message and the `makeRelativeEx` function. To summarize: fsatrace gives absolute paths and `makeRelative` fails to convert to a relative path. Hence shake marks e.g. "/root/a/file.out" (absolute path) as written to by a command, then we may need "../a/file.out" (relative path), but since these strings are not equal (absolute vs relative path) the linting fails to give an error.

DavidEichmann:
Neil, I wanted to add some tests for the new `makeRelativeEx` function, but when I run `cabal new-run shake-test -- filepath` it simply passes without error (even when I e.g. add `True === False`) and says `Warning: No want/action statements, nothing to do`. What am I doing wrong here?

ndmitchell:
Hopefully https://github.com/ndmitchell/shake/blob/master/docs/Developing.md answers your questions. Simple answer is add the word `test` at the end. Just reviewing now...

ndmitchell:
Sorry, I should have clarified what I was looking for. It's clear what `makeRelativeEx` does. What I still don't understand is _why_ GHC is producing files that need it. What GHC rule is it building? What files does it produce? What does it canonicalise them to without makeRelative? How can we write a test case that calls `cmd` and reproduces the error?

DavidEichmann:
### BEFORE this MR
(*) I get the following when building my local GHC branch (includes changes to enable fsatrace linting over the build directory) :
```
$ ./boot && ./configure
$ ./hadrian/build.sh -o../.build/T-16400_default --lint-fsatrace
...
Lint checking error - 36 values were used but not depended upon:
  Used:  /home/david/MEGA/File_Dump/Well-Typed/GHC/_nosync_git/.build/T-16400_default/stage1/libffi/build/configure
  ...
  Used:  /home/david/MEGA/File_Dump/Well-Typed/GHC/_nosync_git/.build/T-16400_default/stage1/libffi/build/missing
  ...
```
In particular, note that I am using a build directory outside of the source tree: `../.build/T-16400_default`. Also note that the lint error gives abosolute paths: the "bad" result of `makeRelative` e.g.:
```
makeRelative "/home/david/MEGA/File_Dump/Well-Typed/GHC/_nosync_git/ghc" "/home/david/MEGA/File_Dump/Well-Typed/GHC/_nosync_git/.build/T-16400_default/stage1/libffi/build/missing"
```
(**) Now if I correctly add a need statement in the corresponding rule (using a relative path as is the convention):
```
need ["../.build/T-16400_default/stage1/libffi/build/missing"]
```
The BUG surfaces: this does **not** fix the linting error. Needing the absolute path does fix the linting error, but is not a satisfactory solution.

### AFTER this MR
With Hadrian as in (*) I do:
```
$ ./boot && ./configure
$ ./hadrian/build.sh -o../.build/T-16400_default --lint-fsatrace
Lint checking error - 31 values were used but not depended upon:
  Used:  ../.build/T-16400_default/stage1/libffi/build/missing
  ...
```
The number of errors decreased from 36 to 31. This is because Hadrian already correctly needs some of the files (using relative paths). Note also that the linting error gives relative paths thanks to `makeRelativeEx`.
Adding the need  as in (**) also fixes the linting error (both relative and absolute paths work).

### To Summarize
fsatrace linting fails when the accessed files are not under the current working directory even if the dependency is coded correctly in the build system (using a relative path).

Does this answer your question?

DavidEichmann:
I've added some test cases now.

ndmitchell:
OK, so the magic ingredient is that you are using a relative path outside the current directory for the build. That's not something I ever really consider - but no reason it shouldn't work. Thanks for the patch!

ndmitchell:
I followed up with a pile of tests and fixes that means it should work robustly now. I'll also write proper docs for this lint mode!

PR #663, opened 2019-03-31
Fix typos

alexfmpe:


ndmitchell:
Thanks a lot! Good catches. I've deployed a new version to the website (might take up to 4 hours to refresh given CDN's)

PR #675, opened 2019-05-01
Add a --compact=auto variant that will not insist in using escape codes unless the terminal will accept them

### LINKED ISSUE #672: Allow to disable --compact
`--compact` is pretty good but it doesn't work in non terminals, such as CI. It's also not possible to set it programmatically. 

We could use a way to disable it from command line (e.g. `--no-compact`), or additional smarts that self-disable it when not in a terminal device ([or when the terminal device is DUMB](http://hackage.haskell.org/package/ansi-terminal-0.9/docs/src/System.Console.ANSI.Unix.html#hSupportsANSI)).

pepeiborra:
Closes #672


ndmitchell:
Thanks! I pushed a few patches on top, to allow --compact=yes|no|auto, and to take the final compact flag (since that's how all the other flags work). Let me know if I broke it for you.

PR #676, opened 2019-05-01
Support GHC 8.8

fumieval:
Pardon an automatically-posted pull request; this is a part of [Operation Vanguard](https://github.com/haskell-vanguard/haskell-vanguard). Note that this does not guarantee that tests and benchmarks are buildable.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Changelog for Shake (* = breaking change)
 
+    Support GHC 8.8
 0.17.9, released 2019-05-01
     #675, allow --compact=yes|no|auto
     Make lintTrackAllow work after the lintTrackRead/Write


ndmitchell:
Thanks, I appreciate the fix! However, this breaks on older GHC versions, which I still support. Any chance you can share the error message? That would give me a clue as to what the compatible fix is likely to be.

fumieval:
This was the error. It looks like it could be an IO exception even an error call.

```
src/Development/Shake/Command.hs:193:9: error:
    • Could not deduce (MonadFail m) arising from a use of ‘fail’
      from the context: MonadTempDir m
        bound by the type signature for:
                   removeOptionFSATrace :: forall (m :: * -> *).
                                           MonadTempDir m =>
                                           Params -> (Params -> m [Result]) -> m [Result]
        at src/Development/Shake/Command.hs:(184,1)-(188,17)
      Possible fix:
        add (MonadFail m) to the context of
          the type signature for:
            removeOptionFSATrace :: forall (m :: * -> *).
                                    MonadTempDir m =>
                                    Params -> (Params -> m [Result]) -> m [Result]
    • In the expression:
        fail
          "Asyncronous process execution combined with FSATrace is not support"
      In an equation for ‘removeOptionFSATrace’:
          removeOptionFSATrace params@Params {..} call
            | not $ isFSATrace params = call params
            | ResultProcess PID0 `elem` results
            = fail
                "Asyncronous process execution combined with FSATrace is not support"
            | otherwise
            = runWithTempDir
                $ \ dir
                    -> do let ...
                          ....
            where
                fsaFlags = fromMaybe "rwmdqt" fsaOptions
                fsaOptions = last $ Nothing : [Just x | FSAOptions x <- opts]
                fsaParams file Params {..}
                  = do prog <- copyFSABinary prog
                       ....
    |
193 |         fail "Asyncronous process execution combined with FSATrace is not support"
```

ndmitchell:
I just pushed a patch to avoid using `fail` in that one place, which is a nicer fix than adding the constraint. Can you try with HEAD?

ndmitchell:
Sorry, an initial bug with the first fix - try again now please.

fumieval:
It works on 8.6 and 8.8. Thanks!

PR #678, opened 2019-05-18
Fix serious bug in writeFile'

mpickering:
This mistake caused a file to be deleted if it had the same name
as the contents of the file you wanted to write.


ndmitchell:
Wow, thanks a lot! Out of interest, how did you find this? Was it hitting something you cared about? I definitely owe you much beer. and probably a dinner!

mpickering:
GHC was using `writeFile'` to write a response file to pass to `ar`. The library I was testing with had only one object file so the file only contained the path to this object file which led to it being deleted just before `ar` was run.

https://gitlab.haskell.org/ghc/ghc/blob/master/hadrian/src/Hadrian/Builder/Ar.hs#L52

ndmitchell:
Thanks @mpickering ! 0.18.2 is now out which fixes this bug.

Issue #680, opened 2019-06-07
on unix shake can only work with ascii or utf8 filenames


typetetris:
From the master branch of shake

 ```
  fileNameToString :: FileName -> FilePath
  fileNameToString = UTF8.toString . fileNameToByteString

  fileNameToByteString :: FileName -> BS.ByteString
  fileNameToByteString (FileName x) = x

  fileNameFromString :: FilePath -> FileName
  fileNameFromString = fileNameFromByteString . UTF8.fromString

  fileNameFromByteString :: BS.ByteString -> FileName
  fileNameFromByteString = FileName . filepathNormalise
```

on unix `fileNameToByteString` is used to convert the filename to a bytestring to be used with
`getFileStatus` and therefore only utf8 or ascii will work here, as the filenames are encoded as utf-8 internally.

That's not problem at all for me personally, but maybe it would be nice to advertise as such on the homepage or in the docs somewhere.

ndmitchell:
Discussion taking place in #681 (TLDR: I don't understand enough to know what I don't know even)

PR #681, opened 2019-06-07
Fix unicode test

typetetris:
Only do the non-ascii test, if the file system encoding
is reported as UTF-8.

Thanks for the pull request!

### DIFF
--- a/src/Test/Unicode.hs
+++ b/src/Test/Unicode.hs
@@ -5,7 +5,6 @@ import Development.Shake
 import Development.Shake.FilePath
 import Test.Type
 import General.GetOpt
-import Control.Exception.Extra
 import Control.Monad
 import GHC.IO.Encoding
 
--- a/src/Test/Unicode.hs
+++ b/src/Test/Unicode.hs
@@ -50,7 +50,7 @@ test build = do
     forM_ ["normal","e^",":)","e^-:)"] $ \pre -> do
         enc <- liftIO getFileSystemEncoding
         if textEncodingName enc /= "UTF-8"
-        then putStrLn $ "WARNING: filesystem encoding is not UTF-8, skipping unicode test (LANG=C ?)"
+        then putStrLn "WARNING: filesystem encoding is not UTF-8, skipping unicode test (LANG=C ?)"
         else do
                 let ext x = decode pre <.> x
                 writeFile (ext "source") "x"
--- a/src/Test/Unicode.hs
+++ b/src/Test/Unicode.hs
@@ -7,6 +7,7 @@ import Test.Type
 import General.GetOpt
 import Control.Exception.Extra
 import Control.Monad
+import GHC.IO.Encoding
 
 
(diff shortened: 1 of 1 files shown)


typetetris:
Why not put that `-Werror=unused-imports` in `shake.cabal`. So I don't need to hit travis first?

ndmitchell:
Thanks for the patch. Sorry for taking so long to get to it - I was travelling and am only just getting back on top of my email. I'll respond to #680 here too.

Generally I use `ghcid` with an appropriate `.ghci` file that turns on all the appropriate warnings. I find that putting warnings in `.cabal` files leads to your users seeing warnings (not what I want).

I thought the whole point about Unix file systems was they didn't really have any encoding, it was all just bytes, with only `/` being special - so you can pack UTF8 in there if you want and it's all happy? As such, what does the encoding of the file system mean? And does anyone use any other type of file names? And doesn't the rule that all Char can be UTF8 mean Shake works with everything? I'm far from an expert at Unix file paths though, being a Windows user, so keen to understand when it fails, where it fails, and if there's any way to stop it failing.

typetetris:
First I want to point out, where shake will fail, if someone uses filenames in a non utf8 encoding.

`getFileStatus` uses `fileNameToByteString`, which assumes via `fileNameFromString`, an utf8 encoding.

In a lot of places `withFile` is used, which boils down to using
```
withFilePath fp f = getFileSystemEncoding >>= \enc -> GHC.withCString enc fp f
```
from `System.Posix.Internals` to convert filepaths from string to bytes, respecting `LC_CTYPE` (or the choice made by `setFileSystemEncoding`).

So the posix filepath (bunch of bytes) which is used to access the file system from those places simply don't match, if a non utf8 encoding is used via `LC_CTYPE`.

Sorry, that I didn't do this presentation in the ticket in the first place.

 > Generally I use `ghcid` with an appropriate `.ghci` file that turns on all the appropriate warnings. I find that putting warnings in `.cabal` files leads to your users seeing warnings (not what I want).

Ah, didn't see that. Thanks for the hint.
 
> I thought the whole point about Unix file systems was they didn't really have any encoding, it was all just bytes, with only `/` being special - so you can pack UTF8 in there if you want and it's all happy?

Correct. But you can also place any other encoding there, if you like. Unix applications usually respect the locale, meaning `LC_CTYPE` for converting file names from strings to bytes.

> As such, what does the encoding of the file system mean?

In my opinion, you just can't assume an encoding for the file system. Either respect the users choice, aka `LC_CTYPE`, or stick to a default and advertise as such.

> And does anyone use any other type of file names?

Really, I can't answer that. In my company we have a strict policy on file names ([a-z_] only, no whitespace), so there is no problem in most locales, as most are compatible on ascii.

> And doesn't the rule that all Char can be UTF8 mean Shake works with everything?

I would assume no. You have to choose an encoding for turning filepaths from strings to bytes and take the correct one, the user also uses for his sources for example.

ndmitchell:
Thanks for the explanation, very useful!

PR #686, opened 2019-07-08
Rename Build.hs to Shakefile.hs

### LINKED ISSUE #582: tutorials: encourage the name “Shakefile.hs”
The online tutorials for shake strangely don’t use the right name (the main project build script should be called `Shakefile.hs` or `Shakefile.lhs` for literate Haskell), which causes the `shake` command to fail in practice because it mainly looks for these two filenames.

Could the tutorials emphasize this filename scheme for a more seamless first time experience?

dwijnand:
Fixes #582

I fixed the filenames and contents without re-running anything so let
me know if anything looks off or if I should verify anything.


dwijnand:
@ndmitchell WDYT?

ndmitchell:
Thanks for the patch. I've been on holiday for the last two weeks and now catching up. Will take a look this evening with luck, and if not, in the next few days.

ndmitchell:
Thanks, much appreciated, and sorry for the huge delay in reviewing. Slowly catching up, but still drowning in PR's and emails!

PR #699, opened 2019-07-27
Use neededMakefileDependencies in the manual

### LINKED ISSUE #696: Shake manual page use need makefile dependencies but probably ment needed makefile dependencies
Hi!

If I understood the documentation correctly the needed variant should be used if the dependencies already was used, which I think is true in the example.

dwijnand:
Fixes #696


ndmitchell:
Thanks for the PR - looks good. Unfortunately it conflicts with the renaming to `Shakefile.hs` - any chance of a revised PR?

ndmitchell:
Thanks! I'll update the website sometime in the next few days.

PR #700, opened 2019-07-27
Add addHelpSuffix, for extra --help info

### LINKED ISSUE #689: suggestion: allow adding 'extra help' to --help output
I have some extra options and target descriptions in my build tool, so the output of `--help` looks like this:

```
./bake.hs --help
...
Extra options:
  --way=WAY                   Set the build 'way'
  --mix=[MIXES...]            Set the build mixes that are included in the 'way'

Targets:
  - clean - Clean all build objects
  - build/debug/src/testing/testing
  - build/debug/src/testing/*.c.o
  - build/debug/src/testing/*.p.h
  - build/debug/src/testing/*.p.o
  - build/release/src/testing/testing
  - build/release/src/testing/*.c.o
  - build/release/src/testing/*.p.h
  - build/release/src/testing/*.p.o
  - all - Build all default targets
  - test - Run the test suite
```

I would like the ability to add *another* section to the help output here, after the `Targets` output, in order to help users a bit more about how to invoke targets and options. Something like:

```
User information:

There are multiple ways to invoke a build...
```

Why do this? In my case, the `--way` and `--mix` options are somewhat complicated to explain inline in the `GetOpt` documentation; notably they can have comma/additive syntax (for instance, `--way=debug,release --mix=asan+lto,ubsan` is a valid incantation that builds the cartesian product of `(Way, [Mix])` variations).

I would like to also indicate to the user things like "what environment variables can impact the build", and so on and so forth, and this seems like a reasonable place to explain it as well. (In theory depending on how it's set up, maybe you could even do things like dump the output of `HashMap String String` for the config files you loaded during the `Rule` phase -- perhaps, so that users can see what build configuration options might be impacting them when they run `--help`)

I suppose simply adding a phony `help` target would also work here, since it would also be listed in the target output under `--help`, so if this seems unnecessary I'm fine with doing that too.

dwijnand:
Fixes #689

This is a lot less polished then I would have preferred, but it took me long enough just to get to this point, with a manual test to verify it works.  Please feel free to point out the silliest of mistakes, as I'm still ramping up in Haskell and there's lots I'm unfamiliar with still.  And advice or guidance is very welcome.

Should I copy `Test.Targets` and build a test that setups up rules with `addHelpSuffix` and test it with `getHelpSuffix`?  Or should I look to be more ambitiuous and create a test for the whole of the `--help` behaviour?

### DIFF
--- a/src/Development/Shake/Internal/Args.hs
+++ b/src/Development/Shake/Internal/Args.hs
@@ -160,15 +160,16 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
     let putWhenLn v msg = putWhen v $ msg ++ "\n"
     let showHelp long = do
             progName <- getProgName
-            targets <- if not long then return [] else
-                handleSynchronous (\e -> do putWhenLn Normal $ "Failure to collect targets: " ++ show e; return []) $ do
+            (targets, helpSuffix) <- if not long then return ([], []) else
+                handleSynchronous (\e -> do putWhenLn Normal $ "Failure to collect targets: " ++ show e; return ([], [])) $ do
                     -- run the rules as simply as we can
                     rs <- rules shakeOpts [] []
                     case rs of
                         Just (_, rs) -> do
                             xs <- getTargets shakeOpts rs
-                            evaluate $ force ["  - " ++ a ++ maybe "" (" - " ++) b | (a,b) <- xs]
-                        _ -> return []
+                            helpSuffix <- getHelpSuffix shakeOpts rs
+                            evaluate $ force (["  - " ++ a ++ maybe "" (" - " ++) b | (a,b) <- xs], helpSuffix)
+                        _ -> return ([], [])
             changes <- return $
                 let as = shakeOptionsFields baseOpts
                     bs = shakeOptionsFields oshakeOpts
@@ -179,7 +180,8 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
                 (if null baseOpts2 then [] else "" : (if null userOptions then "Options:" else "Standard options:") : showOptDescr baseOpts2) ++
                 (if null userOptions then [] else "" : "Extra options:" : showOptDescr userOptions) ++
                 (if null changes then [] else "" : "Changed ShakeOptions:" : changes) ++
-                (if null targets then [] else "" : "Targets:" : targets)
+                (if null targets then [] else "" : "Targets:" : targets) ++
+                (if null helpSuffix then [] else "" : helpSuffix)
 
     when (errs /= []) $ do
         putWhen Quiet $ unlines $ map ("shake: " ++) $ filter (not . null) $ lines $ unlines errs
--- a/src/Development/Shake/Internal/Core/Rules.hs
+++ b/src/Development/Shake/Internal/Core/Rules.hs
@@ -5,13 +5,14 @@
 {-# LANGUAGE TypeFamilies, DeriveDataTypeable #-}
 
 module Development.Shake.Internal.Core.Rules(
-    Rules, runRules,
+    Rules, RulesInfo(..), runRules,
     RuleResult, addBuiltinRule, addBuiltinRuleEx,
     noLint, noIdentity,
     getShakeOptionsRules,
     getUserRuleInternal, getUserRuleOne, getUserRuleList, getUserRuleMaybe,
     addUserRule, alternatives, priority, versioned,
     getTargets, addTarget, withTargetDocs, withoutTargets,
+    addHelpSuffix, getHelpSuffix,
     action, withoutActions
     ) where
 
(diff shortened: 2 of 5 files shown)


src/Development/Shake/Internal/Args.hs
```diff
@@ -169,6 +169,15 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
                             xs <- getTargets shakeOpts rs
                             evaluate $ force ["  - " ++ a ++ maybe "" (" - " ++) b | (a,b) <- xs]
                         _ -> return []
+            helpSuffix <- if not long then return [] else
+                handleSynchronous (\e -> do putWhenLn Normal $ "Failure to get help suffix: " ++ show e; return []) $ do
+                    -- run the rules as simply as we can
+                    rs <- rules shakeOpts [] []
+                    case rs of
+                        Just (_, rs) -> do
+                            helpSuffix <- getHelpSuffix shakeOpts rs
```
ndmitchell: We only want to run the code at most once, so `getHelpSuffix` should be just after `getTargets` above. The code should look alone the lines of `(targets, helpSuffix) <- if not long then return ([], []) else ...`

src/Development/Shake/Internal/Args.hs
```diff
@@ -179,7 +188,8 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
                 (if null baseOpts2 then [] else "" : (if null userOptions then "Options:" else "Standard options:") : showOptDescr baseOpts2) ++
                 (if null userOptions then [] else "" : "Extra options:" : showOptDescr userOptions) ++
                 (if null changes then [] else "" : "Changed ShakeOptions:" : changes) ++
-                (if null targets then [] else "" : "Targets:" : targets)
+                (if null targets then [] else "" : "Targets:" : targets) ++
+                (if null helpSuffix then [] else "" : "User information:" : [helpSuffix])
```
ndmitchell: I wouldn't bother prepending `User information` - the user can add that with the suffix if they want.

src/Development/Shake/Internal/Core/Rules.hs
```diff
@@ -160,14 +166,15 @@ data SRules = SRules
     ,builtinRules :: !(Map.HashMap TypeRep{-k-} BuiltinRule)
     ,userRules :: !(TMap.Map UserRuleVersioned)
     ,targets :: !(ListBuilder Target)
+    ,helpSuffix :: !String
```
ndmitchell: This should be a `ListBuilder String` so we never get into quadratic append problems.

src/Development/Shake/Internal/Core/Rules.hs
```diff
@@ -133,12 +134,12 @@ modifyRulesScoped f (Rules r) = Rules $ do
         modifyIORef' refOld (<> f rules)
         return res
 
-runRules :: ShakeOptions -> Rules () -> IO ([(Stack, Action ())], Map.HashMap TypeRep BuiltinRule, TMap.Map UserRuleVersioned, [Target])
+runRules :: ShakeOptions -> Rules () -> IO ([(Stack, Action ())], Map.HashMap TypeRep BuiltinRule, TMap.Map UserRuleVersioned, [Target], String)
```
ndmitchell: Now there are 4 fields it probably wants to be a real return data type, e.g. `data RulesInfo = {riActions :: [(Stack, Action ()]], ...}`. However, this isn't essential.

ndmitchell:
Thanks a lot! I'd probably just add a quick little test to `Test.Targets` rather than copying it. I don't imagine `addHelpSuffix` is going to be a serious source of bugs. If you do want to do something fancy with testing `--help` that would be most welcome, but certainly not essential at this stage.

dwijnand:
The macOS build failed with:

```
error: RPC failed; curl 18 transfer closed with outstanding read data remaining
fatal: The remote end hung up unexpectedly
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-cask failed!
```

I heard about this yesterday, there's something going on with homebrew and GitHub: https://discourse.brew.sh/t/fetching-homebrew-repos-is-slow/5374

src/Development/Shake/Internal/Core/Rules.hs
```diff
@@ -147,9 +148,14 @@ runRules opts (Rules r) = do
 --   Returns the command, paired with the documentation (if any).
 getTargets :: ShakeOptions -> Rules () -> IO [(String, Maybe String)]
 getTargets opts rs = do
-    (_actions, _ruleinfo, _userRules, targets) <- runRules opts rs
+    RulesInfo{riTargets = targets} <- runRules opts rs
```
ndmitchell: I'd have gone with record punning and just written `RulesInfo{riTargets}` and used `riTargets` on the line below. However, its basically personal style.

ndmitchell:
Thanks!

ndmitchell:
After this change it was pretty clear that `RulesInfo` and `SRules` were really the same thing with just lists instead of listbuilders. As a result, I generalised them, so now they are actually the same with a higher-kinded argument type. Thanks for doing the cleanup that made that clear.

PR #702, opened 2019-08-05
Sync docs/Manual.md with docs/manual/Shakefile.hs

dwijnand:


ndmitchell:
Thanks! `cmd_` didn't exist when this was first written :)

PR #703, opened 2019-08-06
Print rebuild warnings in yellow & errors in red

### LINKED ISSUE #640: Choose a better color when Shake warns about database rule changes
When database rules change, Shake will rerun rules affected by that. However, when using `--color` to make the output pretty, *everything* is printed in blue, including my rules and the warning:

```shell
./shake.hs --color
Shake database rules have changed for the following types:
  File:  .shake/.shake.database
  Type:  OracleQ Config {7daca85a005f975e}
  Type:  OracleQ ConfigKeys {3cf7c26d5ae5e884}
All rules using these types will be rebuilt

...
```

This is pretty easy to miss, especially if your build has a lot of rules you didn't expect to rebuild. It would be nice if this was a distinct color in the terminal output (for example, yellow.)

dwijnand:
Builds on #706
Fixes #640

---
Original description:

I'm submitting this for early feedback.  (https://42floors.com/blog/startups/thirty-percent-feedback)

This is what it looks like in an out-of-repo, manual test I have:

image

But I suspect that duplicating `shakeOutput` is not a good strategy, so I'm looking for some advice.

If this way is OK, then I'll see to adding a test for it, adding it to the release notes, and adjusting the docs (around its introduction with respects to `shakeOutput`.)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,5 +1,6 @@
 Changelog for Shake (* = breaking change)
 
+    #703, print rebuild warnings in yellow and errors in red
     #706, rename Verbosity's constructors and add Warn
     #443, add getEnvError
     #662, add Partial to command/cmd
--- a/src/Development/Shake/Internal/Args.hs
+++ b/src/Development/Shake/Internal/Args.hs
@@ -294,7 +294,12 @@ escape :: Color -> String -> String
 escape color x = escForeground color ++ x ++ escNormal
 
 outputColor :: (Verbosity -> String -> IO ()) -> Verbosity -> String -> IO ()
-outputColor output v msg = output v $ escape Blue msg
+outputColor output v msg = output v $ color msg
+  where color = case v of
+            Silent -> id
+            Error  -> escape Red
+            Warn   -> escape Yellow
+            _      -> escape Blue
 
 -- | True if it has a potential effect on ShakeOptions
 shakeOptsEx :: [(Bool, OptDescr (Either String ([Extra], ShakeOptions -> ShakeOptions)))]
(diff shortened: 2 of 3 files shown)


thoughtpolice:
Whoo! Thank you for these PRs to fix my (simple) issues, @dwijnand!

dwijnand:
No worries, keep them coming!  I'll try to take care of the easy pickings as I ramp up.

ndmitchell:
Thanks, and I appreciate you sharing it early.

At the moment if people want to hook messages they just hook one thing, which I prefer. If you have output warning, why not also have output error? And it seems highly intertwined with verbosity levels - should there be specific error/warning levels in verbosity? Is the verbosity thing pretty dumb anyway and there should be more standard logging levels of `Debug | Warn | Err | Info` ? This is an area where there's already too much complexity, and too many people get confused by the `Verbosity`, so I'm not averse to a more significant overhaul. See #369 for some of the confusion, and whenever I try and describe the way it works, I often get it backwards...

dwijnand:
So I looked at some Shake code, issues and PRs, and compared it to some prior art within the industry, and I suggest the following:
1. We (further) associate a log _level_ to the constructors of `Verbosity`, either indirectly with documentation or directly by renaming them (e.g. rename `Quiet` to `Error`)
2. Introduce a `Warn` or `Warning` constructor to `Verbosity`, that sits between `Quiet`/`Error` (and associated `putWarn`, etc functions)
3. We colour according to log level, so warnings are yellow, errors become red, normal output stays blue, etc

WDYT?

ndmitchell:
Yep, I like that. Furthermore we can use pattern synonyms to keep "compatible" names so we can phase out Loud but keep it for compatibility. Keen to see a fully sketched design - just something like the proposed Verbosity data type and functions. Having screwed up the design once on my own keen to have a few eyes on this one!

dwijnand:
A much smaller diff!

I'll rotate back to this to complete it.

dwijnand:
Opened #706 for changing Verbosity, and rebased this level/colour change on top of it.

dwijnand:
Green modulo "Ninja build was faster than Shake".

ndmitchell:
GitHub sucks for stacked diffs. A quick skim looks great, but we should land #706 first.

ndmitchell:
Can you rebase somehow? Otherwise I'll struggle to review this patch at all (I blame GitHub...)

dwijnand:
Rebased.

ndmitchell:
Wow, the rebased change is so small. Thanks a lot!

PR #704, opened 2019-08-08
Apply a HLint suggestion to a doc example

dwijnand:
HLint's output was:

    ./Shake.hs:25:44: Suggestion: Use <$>
    Found:
      fmap fromStdout $ cmd "ghc --numeric-version"
    Perhaps:
      (fromStdout Control.Applicative.<$> cmd "ghc --numeric-version")

    1 hint

### DIFF
--- a/src/Development/Shake/Internal/Rules/Oracle.hs
+++ b/src/Development/Shake/Internal/Rules/Oracle.hs
@@ -82,7 +82,7 @@ addOracleFlavor flavor act = do
 -- newtype GhcVersion = GhcVersion () deriving (Show,Typeable,Eq,Hashable,Binary,NFData)
 -- type instance RuleResult GhcVersion = String
 -- rules = do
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks, good idea. Do you have to \< and \> to escape the output? I think you tripped Haddock markup, which is why the test is failing.

dwijnand:
This failed oh GHC 8.6 with

```
Ninja was 10821 then 5
Shake was 11143 then 13
travis: ERROR: Ninja build was faster than Shake
```

which I think is a flakiness in the test (maybe?).  Other than that it's good.

ndmitchell:
That test is a performance test, comparing the speed of Ninja vs Shake, so it has a bit of flakiness on a CI machine unfortunately.

ndmitchell:
And thanks for the patch!

PR #705, opened 2019-08-11
Fix links to the "apply" function.

dwijnand:


ndmitchell:
Thanks! I do wish Haddock had an automated way to spot such problems and fail CI, but alas, I haven't found one.

PR #706, opened 2019-08-11
Rename Verbosity's constructors and add Warn

dwijnand:
Based off of https://github.com/ndmitchell/shake/pull/703#issuecomment-519429768.

I've picked some names based on the ones I know from the JVM ecosystem as well as looking at syslog's severity levels.

`Error` is already used elsewhere within Shake so I went with `Err` (which I'm not loving).

Also, before ending up with `Verbose` I originally had `Trace` and then `Traced`, but both of those are also taken.

I won't lie, the fact that the new names go from short to long the more verbose the logs get (Info -> Debug -> Verbose -> Diagnostic) is really pleasing me, but I'm a simple man.

I've left `quietly` as I didn't know what to use instead with the new associated Verbosity constructor being `Error` - ideas?

Also how should `-q`/`--quiet` and `-V`/`--verbose`/`--trace` be dealt with... hmm...

Fixes https://github.com/ndmitchell/shake/issues/369


dwijnand:
OK, this is green now (outside of "Ninja build was faster than Shake" errors).

CHANGES.txt
```diff
@@ -4,6 +4,7 @@ Changelog for Shake (* = breaking change)
     #689, allow adding help messages at the end using addHelpSuffix
     #686, rename Build.hs to Shakefile.hs
     #690, make Ninja mode use -j0 by default (more like Ninja)
+    #706, rename Verbosity's constructors and add Warn
```
ndmitchell: I usually put these things at the top of the changelog, FWIW - however, it doesn't really matter and maybe this is a better way to avoid conflicts. However, growing up makes it easier to see the diff between prereleases.

src/Development/Shake.hs
```diff
@@ -67,7 +67,10 @@ module Development.Shake(
     -- ** Progress reporting
     Progress(..), progressSimple, progressDisplay, progressTitlebar, progressProgram, getProgress,
     -- ** Verbosity
-    Verbosity(..), getVerbosity, putLoud, putNormal, putQuiet, withVerbosity, quietly,
+    Verbosity(..), getVerbosity, putDebug, putInfo, putErr, withVerbosity, quietly,
+#if __GLASGOW_HASKELL__ < 800
+    pattern Quiet, pattern Normal, pattern Loud, pattern Chatty,
```
ndmitchell: Why aren't these patterns defined in this module too?

src/Development/Shake/Internal/Core/Action.hs
```diff
@@ -158,25 +158,25 @@ putWhen v msg = do
...
-putNormal = putWhen Normal
+putInfo :: String -> Action ()
+putInfo = putWhen Info
 
 -- | Write an important message to the output, only supressed when 'shakeVerbosity' is 'Silent'.
 --   The output will not be interleaved with any other Shake messages (other than those generated by system commands).
-putQuiet :: String -> Action ()
-putQuiet = putWhen Quiet
+putErr :: String -> Action ()
+putErr = putWhen Err
```
ndmitchell: Agreed, `putError` would be much better!

src/Development/Shake/Internal/Options.hs
```diff
@@ -301,14 +306,24 @@ tyHidden = mkDataType "Development.Shake.Types.Hidden" []
 
 -- | The verbosity data type, used by 'shakeVerbosity'.
 data Verbosity
-    = Silent -- ^ Don't print any messages.
-    | Quiet  -- ^ Only print essential messages, typically errors.
-    | Normal -- ^ Print errors and @# /command-name/ (for /file-name/)@ when running a 'Development.Shake.traced' command.
-    | Loud   -- ^ Print errors and full command lines when running a 'Development.Shake.command' or 'Development.Shake.cmd' command.
-    | Chatty -- ^ Print errors, full command line and status messages when starting a rule.
+    = Silent  -- ^ Don't print any messages.
+    | Err     -- ^ Only print error messages.
```
ndmitchell: Definitely `Error`.

src/Development/Shake/Internal/Options.hs
```diff
@@ -301,14 +306,24 @@ tyHidden = mkDataType "Development.Shake.Types.Hidden" []
...
+    | Warn    -- ^ Print errors and warnings.
+    | Info    -- ^ Print errors, warnings and @# /command-name/ (for /file-name/)@ when running a 'Development.Shake.traced' command.
+    | Debug   -- ^ Print errors, warnings and full command lines when running a 'Development.Shake.command' or 'Development.Shake.cmd' command.
+    | Verbose -- ^ Print errors, warnings, full command line and status messages when starting a rule.
     | Diagnostic -- ^ Print messages for virtually everything (mostly for debugging).
       deriving (Eq,Ord,Show,Read,Typeable,Data,Enum,Bounded)
 
+-- | A bidirectional pattern synonym for 'Err'.
+pattern Quiet  = Err
```
ndmitchell: I'd move these into `Development.Shake` as these are effectively on the deprecation path now.

dwijnand: From https://gitlab.haskell.org/ghc/ghc/wikis/pattern-synonyms/associating-synonyms I had understood that defining them here keeps them bundled with `Verbosity` so, for recent enough versions of GHC, if you import `Verbosity(..)` you get the old patterns and constructors too.

Maybe that isn't true but, either way, it's probably unnecessary, so I moved them as you suggested.

dwijnand:
> Only minor thought I had is having both `Debug` and `Verbose` is a bit confusing. They aren't both standard levels in logging. I wonder about getting rid of `Verbose` entirely. I'm not sure it's that useful, and the fact it has no good name makes it even less useful.

`--verbose` is standard for CLI utilities and `debug` is a log level in JVM logging libraries.  But together, not so much.  One alternative is using `Fine`, `Finer`, or `Finest` from [Java](https://docs.oracle.com/javase/8/docs/api/java/util/logging/Level.html).  Or we could rename `Trace` in Shake to `TraceEvent` or `TraceMessage` and rename `Verbose to `Trace.

ndmitchell:
I didn't realise you could put the `putDebug` Haddock comments where you did - I thought they had to be on the Haddock. I learned something today.

I went ahead and marks the pattern synonyms as deprecated, since people should use the real patterns now.

I'm still a bit sad to have both debug and verbose, since they seem basically the same thing. There is no `putVerbose` and only two places were print things if you have verbose set - namely when starting a file rule. I'm inclined to collapse Verbose and Debug, and probably print out the message when starting a file rule on Debug as well. That way we align more closely with existing approaches. @dwijnand  - thoughts?

dwijnand:
No strong feeling about it from me (not enough info to be able to tell).  Dropping the very similar "debug" and "verbose" is wise.  Oh, I guess merging them means that previous `-vv` would become `-vvv` (Diagnostic) but I'm guessing (hoping) that no one's scripted around `-vv` usage...

ndmitchell:
I deleted Debug in 14ae18ac58e87a9573fff54cf080acd704e2b570 - I'd appreciate a quick sanity check of that patch to confirm it's going in a direction you consider consistent with the excellent work you've done.

dwijnand:
Looks good to me.

Perhaps it's overkill to have the deprecated `Debug` and `putDebug` aliases, given we have yet to ship those.

ndmitchell:
Agreed, I hadn't realised we hadn't shipped them - I've deleted them now.

dwijnand:
Nice work, the changes look good

PR #708, opened 2019-08-13
Rename "deprioritize" to "reschedule'

### LINKED ISSUE #629: priority vs deprioritise is confusing

dwijnand:
... and rename "preference" back to "priority".

Fixes #629 (again)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,6 @@
 Changelog for Shake (* = breaking change)
 
-    #629, rename priority to preference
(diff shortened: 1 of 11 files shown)


ndmitchell:
Thanks!

PR #709, opened 2019-08-16
Add Partial to command/cmd

### LINKED ISSUE #662: command/cmd do not have partial annotations
Mean you don't get stack traces from them, which would be pretty handy

dwijnand:
Fixes #662


ndmitchell:
Thanks for the patch! Unfortunately it isn't just as simple as adding Partial - that doesn't give you call stacks. You have to propagate the Partial to where error is called, make sure you call the right sort of error, and then mask out the internal details. It's quite a fiddly and involved process, and most of the time you also need to eta expand something or it doesn't work. I played around and put a few patches on top, which gives us call stacks on cmd/command.

dwijnand:
I see - I had assumed that it would've errored if the propagation wasn't there.  I'll read the commit log to see the long tail of the necessary changes.

ndmitchell:
Lack of error propagation just means the error stops propagating - it's a redundant constraint. Turning on `-Wredundant-constraint` does seem to highlight some of them. Unfortunately it also highlights things where shake uses the RuleResult constraint, things which use Partial more as a tag than for errors (e.g. writeFileChanged), things like Unit on cmd_ etc. It's also not available on GHC 7.10 (the oldest shake supports).

PR #712, opened 2019-08-17
Add a partial variant of getEnv

### LINKED ISSUE #443: Add partial version of getEnv
This is a suggestion to add a partial version of `getEnv`, because I end up defining it in almost every shake build system; maybe others, too?

Something like

``` haskell
getEnv_ :: String -> Action String
getEnv_ name =
  getEnvWithDefault
    (error $ "Environment variable " ++ name ++ " is undefined")
    name
```

Where `error` probably should be replaced by a more suitable error signalling mechanism.

dwijnand:
Fixes #443

### DIFF
--- a/src/Development/Shake.hs
+++ b/src/Development/Shake.hs
@@ -92,7 +92,7 @@ module Development.Shake(
     doesFileExist, doesDirectoryExist, getDirectoryContents, getDirectoryFiles, getDirectoryDirs,
     getDirectoryFilesIO,
     -- * Environment rules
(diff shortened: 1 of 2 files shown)


ndmitchell:
Thanks! Good choice of name.

dwijnand:
A wise man [suggested it](https://github.com/ndmitchell/shake/issues/443#issuecomment-214788416) 😄

ndmitchell:
I had totally forgotten that!

PR #713, opened 2019-08-17
Travis CI: Switch to "language: generic"

dwijnand:
Switches awway from the default, which is actually "language: ruby". See
https://docs.travis-ci.com/user/languages/minimal-and-generic/

From some testing this looks like it shaves 4 minutes off the total
build, from 1h51m to 1h47m.  Not amazing, but still net positive.

Also drop the "sudo: required" that now is a no-op.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: generic
+language: minimal
 matrix:
   include:
   - env: GHCVER=7.10
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: required
+language: generic
 matrix:
   include:
   - env: GHCVER=7.10


ndmitchell:
When I last tried removing `sudo: required` there was a massive performance overhead - like 2x in some cases. But that was many years ago, I'll give it a whirl.

I also see `language: minimal` that looks like enough for my projects? With these more general CI changes I usually test them on my `neil` project (https://github.com/ndmitchell/neil) and then roll them out to everyone, so I'll do that first. The changes look like a great idea, and while times vary quite a bit, using the morally right thing in generic/minimal makes me happy.

dwijnand:
Yeah the `sudo: required` and `sudo: false` thing (aka Virtual-Machine-Based vs. Container-Based Builds), died last year: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration.  If I PR a .travis.yml I boy scout it 😄

I had tried the `minimal` base (looks like Travis CI has deleted my logs? https://travis-ci.com/dwijnand/shake/builds/123558298), but it failed.  I believe it was on the `npm` call in the repo-local `travis.hs`... but now I see an apt-get install` just above it, which has me confused.

I'll push that diff so we can see the failure together here.

ndmitchell:
I'm debugging this on top of your patch against the server.

dwijnand:
Looks like either add:

```bash
sudo apt-get install npm
sudo ln -s /usr/bin/nodejs /usr/bin/node
```

(from https://www.hostingadvice.com/how-to/install-nodejs-ubuntu-14-04/#ubuntu-package-manager)

Or, alternatively (and looks like more officially?): https://nodejs.org/en/download/package-manager/#debian-and-ubuntu-based-linux-distributions-enterprise-linux-fedora-and-snap-packages

ndmitchell:
I added both of those lines you suggested and then minimal works (I also switched the rest of my Travis files to minimal with great success). There was some fallout in that tslint stopped working (see #717) but that's not a big deal (tslint is useful when I'm actively hacking on the profile stuff, but that's sporadic).

I am unconvinced there are any measurable speed benefits (the times across all my builds are roughly the same), but it's still good hygiene. Thanks!

PR #714, opened 2019-08-17
Use Data.Bifunctor.second (hlint)

dwijnand:
./src/Development/Shake/Internal/Core/Database.hs:110:63: Suggestion: Use second
    Found:
      \ (k, s) -> (k, f s)
    Perhaps:
      Data.Bifunctor.second f
    Note: increases laziness


ndmitchell:
Thanks a lot!

Issue #716, opened 2019-08-20
CI Instability


ndmitchell:
The Ninja slower than Shake test is now failing in CI with reasonable frequency. Did Ninja get faster? Shake get slower? Or some other variable?

PR #718, opened 2019-08-27
Make shakeSymlink=False the default.

TravisWhitaker:
This simply swaps the `shakeSymlink` default to `False`. With the current behavior, something like this is bound to happen to users who try this feature:

- `shake ... --share=...`
- Edit a file.
- `shake ... --share=...`
- `error: unable to open output file 'build/src/foo.o': 'Operation not permitted'`

Using symlinks by default probably makes sense for CI jobs or other robots that run builds, but those robots can easily use `--share-symlink` when building. `--share-copy` should be the default for a person working in a source tree.


dwijnand:
Yeah https://github.com/ndmitchell/shake/issues/716

TravisWhitaker:
@ndmitchell curious about your thoughts on this.

TravisWhitaker:
Pinging recent committers @ndmitchell @dwijnand @adamse

adamse:
My thoughts:

- Copy might be an easier default since you don't have to modify your buildsystem to delete output files (I think the error is due to setting readonly on the links to not corrupt the cache)
- This might be a breaking change, so should be marked that way in the changelog?
- As far as I understand shake is using hard-links (or maybe it's called just links), calling them symlinks is confusing
- You can change the default for your buildsystem by passing `shakeOptions{shakeSymlink=False/True}` to https://hackage.haskell.org/package/shake-0.18.3/docs/Development-Shake.html#v:shakeArgs

ndmitchell:
Thanks for all the thoughts! Agreed that the symlink+RO has to go together, but that makes it a less pleasant solution.

It is technically a breaking change, but not a code-breaking change, and in a part that is brand new and not that stable. I've opted to mark it with a * in the changelog, but not give a 0.1 bump.

I also noted that these are actually hard links not sym links in the docs. I'm mostly just annoyed that file systems are copy-on-write underneath, and yet that's not in the POSIX API so no one exposes it. I also documented it all on the Cloud.md website page.

PR #719, opened 2019-09-02
Write cache key after cached files to ensure cache is sane

adamse:
It might happen that we are interrupted before writing all files, by writing
the key after the files we ensure we don't get a key with missing files.


ndmitchell:
Thanks! Very sensible.

PR #721, opened 2019-09-05
Handle GHC 8.8 and MonadFail

dwijnand:
This is the tail end changes of the MonadFail proposal.

### DIFF
--- a/src/Development/Shake/Internal/Core/Database.hs
+++ b/src/Development/Shake/Internal/Core/Database.hs
@@ -19,7 +19,7 @@ import Control.Concurrent.Extra
 import Control.Monad.IO.Class
 import qualified General.Ids as Ids
 
-#if __GLASGOW_HASKELL__ >= 800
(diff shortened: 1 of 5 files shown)


ndmitchell:
I see a weird test failure pop up.

```
TESTS FAILED
shake-test: ASSERTION FAILED: File contents are wrong: catch1
WANT: magic1
GOT: user error (magic1)
CallStack (from HasCallStack):
  error, called at src/Test/Type.hs:158:18 in main:Test.Type
Test suite shake-test: FAIL
```

Is this a consequence of this patch? It's on the GHC 8.8 branch, around the area of exceptions being thrown, so could potentially have shifted something? That said, having read the patch it looks very good and obviously correct. It was failing earlier before, so this is strictly better, and I'll merge it.

ndmitchell:
Oh, and always but I forgot to say this time, Thanks!!!

dwijnand:
I looks like it might be, and probably worth revisiting before cutting a release that supports 8.8.

PR #724, opened 2019-09-07
Decrease the leeway given to shake vs ninja

dwijnand:
For #716


ndmitchell:
Thanks, a combination of two hacks combining and fighting. I think this ends up actually increasing the leeway :). I had the `* 2` on the wrong side the first time 🤦‍♂

dwijnand:
Oh yeah, haha, I hadn't noticed that.

Issue #722, opened 2019-09-05
cmd and cmd_ get slow on linux if there is a high RLIMIT_NOFILES (open files)


typetetris:
https://github.com/ndmitchell/shake/blob/0eda14aaea8a893524248855064a3415ab3fe67a/src/General/Process.hs#L155

Here `close_fds = True` is set, which causes https://hackage.haskell.org/package/process-1.6.5.1/docs/System-Process.html#v:createProcess to loop from 3 to the maximum file descriptor number the process might have (RLIMIT_NOFILES, ulimit -n) and calling `close` on each one.
(Can be easily confirmed by using `strace`.)

In an environment with `ulimit -n` set to 1000000, this burns 10 seconds on around 30 `cmd_`
calls on a machine used at the company I work for. (Small demo shake system, only doing
 `touch`.)

If you don't know what is going on, you have that puzzling behaviour of having a very slow build system on one machine/environment and a fast one on a different machine/environment without a clue, why that is. Even knowing, what is going on now, it is still a very surprising behaviour.

Please consider making `close_fds` configurable and change the default to `False` (at least on linux). That's the usual behaviour of most programs shelling out commands, like bash, make, etc ..

If you explain, how would like to handle that, I am glad to provide a PR.

ndmitchell:
That's horrible! And very hard to discover - well done for finding it. I set `close_fds` because in #243 @nomeata said I should. Given the performance impact, I'm inclined to set it back to `False`. However, I'm woefully ignorant here, so I'll just follow everyone else's advice 😄.

typetetris:
I suggest making `close_fds` configurable (with an parameter to `cmd_` like `Shell`.) and setting the default to `False`. I suggest `CloseFDs` as an identifier.

Of course, that's an API change, which needs to be advertised, at least in the change notes.

The database should be opened with a O_CLOEXEC flag (I need to research, how to do this with the haskell standard library.) That would address #243 a bit.

Some arguments:

* If one kept the default `close_fds = True`, the behaviour needs to be explained in the manual directly in the documentation of `cmd` and `cmd_`, as it is not the expected behaviour and has potential performance implications. The other way round, you only need explanations on a strange, rarly used flag `CloseFDs`. 

* On linux, you usually flag fds with FD_CLOEXEC which you don't want to inherit to shelled out commands, so the OS closes them for you efficiently on any exec* call. (But that's a consideration
more for the standard library.)

* Its the default behaviour of most unix programs shelling out commands.

[...]

PR #727, opened 2019-09-15
#722, Add CloseFileHandles and default to not using it

typetetris:
On linux closing unrelated file handles isn't the expected/supported
behaviour. The implementation in the used library to this was
slow too, as it iterates over every possible open file descriptor number.

There are two places in the provided test case, where I would be glad for different
suggestions. How can I do a `bracket` like thing in `Action ()`? For the test case
its probably not a problem, but I would love to know.

Also I don't know, how all of that would be tested on windows. Sorry for that.

Thanks for the pull request!

### DIFF
--- a/src/Development/Shake/Internal/CmdOption.hs
+++ b/src/Development/Shake/Internal/CmdOption.hs
@@ -28,4 +28,5 @@ data CmdOption
     | AutoDeps -- ^ Compute dependencies automatically.
     | UserCommand String -- ^ The command the user thinks about, before any munging. Defaults to the actual command.
     | FSAOptions String -- ^ Options to @fsatrace@, a list of strings with characters such as @\"r\"@ (reads) @\"w\"@ (writes). Defaults to @\"rwmdqt\"@ if the output of @fsatrace@ is required.
+    | CloseFileHandles -- ^ Before starting the command in the child process, close all file handles except stdin, stdout, stderr in the child process. Uses @close_fds@ from package process and comes with the same caveats, i.e. runtime is linear with the maximum number of open file handles (RLIMIT_NOFILE, see man 2 getrlimit on linux).
       deriving (Eq,Ord,Show,Data,Typeable)
--- a/src/Test/CloseFileHandles.hs
+++ b/src/Test/CloseFileHandles.hs
@@ -1,0 +1,69 @@
+{-# LANGUAGE CPP #-}
+module Test.CloseFileHandles(main) where
+
+import Development.Shake
+import Development.Shake.FilePath
+import Test.Type
+import Control.Monad.Extra
+import System.Exit
+import System.IO
+
+#ifdef mingw32_HOST_OS
+
+main = testNone -- don't know how to do this on windows
+
+#else
+
+import System.Posix.IO
+
+main = testBuild test $ do
+    let helper = toNative $ "helper/close_file_handles_helper" <.> exe
+    let name !> test = do want [name]
+                          name ~> do need ["helper/close_file_handles_helper" <.> exe]; test
+
+    let helper_source = unlines
+            ["import System.Environment"
+            ,"import System.Posix.IO"
+            ,"import System.IO"
+            ,"import System.Exit"
+            ,""
+            ,"main = do"
+            ,"  args <- getArgs"
+            ,"  case args of"
+            ,"    [fdString] -> do"
+            ,"       handle <- fdToHandle (read fdString)"
+            ,"       hClose handle"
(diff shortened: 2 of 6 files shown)


typetetris:
aww  `fail` is probably not the correct thing to do, will reformulate to same assert call.

src/Development/Shake/Internal/CmdOption.hs
```diff
@@ -28,4 +28,5 @@ data CmdOption
     | AutoDeps -- ^ Compute dependencies automatically.
     | UserCommand String -- ^ The command the user thinks about, before any munging. Defaults to the actual command.
     | FSAOptions String -- ^ Options to @fsatrace@, a list of strings with characters such as @\"r\"@ (reads) @\"w\"@ (writes). Defaults to @\"rwmdqt\"@ if the output of @fsatrace@ is required.
+    | CloseFileHandles -- ^ Before starting the command in the child process, close all file handles except stdin, stdout, stderr in the child process. Uses close_fds from package process and comes with the same caveats, i.e. runtime is linear with the maximum number of open file handles (RLIMIT_NOFILE).
```
ndmitchell:
```suggestion
    | CloseFileHandles -- ^ Before starting the command in the child process, close all file handles except stdin, stdout, stderr in the child process. Uses @close_fds@ from package process and comes with the same caveats, i.e. runtime is linear with the maximum number of open file handles (@RLIMIT_NOFILE@).
```

src/Test/CloseFileHandles.hs
```diff
@@ -0,0 +1,71 @@
...
+    "close_file_handles_helper.hs" %> \out -> do
+        need ["../../src/Test/CloseFileHandles.hs"]
+        writeFileChanged out helper_source
+    ["helper/close_file_handles_helper"<.>exe, "close_file_handles_helper.hi", "close_file_handles_helper.o"] &%> \_ -> do
+        need ["close_file_handles_helper.hs"]
+        cmd "ghc --make" "close_file_handles_helper.hs -o helper/close_file_handles_helper" 
+
+    "defaultbehaviour" !> do
+        withTempFile $ \file -> do
+            h <- liftIO $ openFile file AppendMode  -- should probably be a bracket in Action () how to do that?
```
ndmitchell: It was impossible. I just added `actionBracket` that should be perfect for this. Can you try converting the code over and report back?

src/Test/CloseFileHandles.hs
```diff
@@ -0,0 +1,71 @@
...
+            fd <- liftIO $ handleToFd h
+            (Exit c, Stdout _, Stderr _) <- cmd "helper/close_file_handles_helper" (show fd) :: Action (Exit, Stdout String, Stderr String)
+            liftIO $ hClose h
+            liftIO $ assertBool (c == ExitSuccess) "handle closed without option CloseFileHandles" 
+
+    "closing" !> do
+        withTempFile $ \file -> do
+            h <- liftIO $ openFile file AppendMode  -- should probably be a bracket in Action () how to do that?
```
ndmitchell: Ditto, there is now actionBracket.

src/Test/CloseFileHandles.hs
```diff
@@ -0,0 +1,71 @@
...
+import Development.Shake
+import Development.Shake.FilePath
+import Test.Type
+import Control.Monad.Extra
+import System.Exit
+import System.IO
+
+#ifdef mingw32_HOST_OS
+
+main = return () -- don't know how to do this on windows
```
ndmitchell: This is the wrong type. Use `testNone` (which I just added).

ndmitchell:
Btw, `fail` is entirely fine in a test. The assertion things are just helpers, not mandatory.

typetetris:
Fixed up a bit.

typetetris:
Ok, looks like doing `@RLIMIT_NOFILES@` breaks some doc tests.

What can I do? Should I just remove those?

dwijnand:
I see it failing for

```
Defined but not used: ‘helper’
```

typetetris:
Yes, that's the Travis failure. I looked at the AppVeyor thing.

typetetris:
Now both failures are because of the line
`code_58 = RLIMIT_NOFILE`
in some generated code.

Maybe I just remove the `@` for now.

ndmitchell:
Thanks! FWIW, I fixed the docs by doing c2ea9e91e93f69bc0af6d4f99b3f7f5641c8344b - experience tells me that most code in docs is wrong by default, so I default to assuming its valid Haskell, but then have to whitelist what isn't.

PR #736, opened 2019-12-30
Fix typos in Index.md

srid:
Thanks for mentioning rib! :-)


ndmitchell:
Thanks! I've updated the website (may take up to 4 hours to so to trickle through)

PR #741, opened 2020-01-22
Documented new use of cmd after IsCmdArgument change.

jpcooper:


jpcooper:
@ndmitchell Given that `cmd` doesn't only create `Action`s, should we remove `:-> Action r` in `args :-> Action r` as well?

ndmitchell:
Thanks! It was never really true that `cmd` returns an `Action` - you can always run it in `IO` as well. It's more a hint as to what kind of things people should be using it for, and as that hint, I imagine it's going to 95%+ be used for `Action` - so it seems to make sense to keep the hint.

jpcooper:
Fair enough.

PR #742, opened 2020-01-28
Filter to latest build run in the chrome profiles

nickhs-asana:
Previously all information from the build database was included
in chrome profiles without any delineation between build runs.
This made chrome profiles difficult to comprehend.

This change now only writes the latest build run information
to the chrome profile.

Thanks for the pull request!

### DIFF
--- a/src/Development/Shake/Internal/Profile.hs
+++ b/src/Development/Shake/Internal/Profile.hs
@@ -94,7 +94,6 @@ data ProfileTrace = ProfileTrace
     {prfCommand :: String, prfStart :: Double, prfStop :: Double}
 prfTime ProfileTrace{..} = prfStop - prfStart
 
-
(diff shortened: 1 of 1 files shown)


nickhs-asana:
Hi @ndmitchell I didn't open an issue per https://github.com/ndmitchell/neil#contributions but do let me know if you'd prefer to discuss in there.

This is a small change that attempts to make the chrome profile trace more useful by only writing the latest build run to the trace file, instead of all possible builds. The latter makes the profiles hard to read where each run is overlapping each other.

Apologies in advance if I've misunderstood something fundamental about how the chrome profile tracing was meant to work.

ndmitchell:
Thanks, this is great. Having multiple overlapping profiles in a chrome trace makes it very hard to see anything. Good catch!

PR #749, opened 2020-03-09
Don't hang when running an action via Docker (fixes #748)

### LINKED ISSUE #748: Docker invocation hangs starting at Shake 0.18.3
```
import Development.Shake

main :: IO ()
main = shakeArgs shakeOptions{shakeFiles = "_build"} $ do
    phony "foo" $ do
        cmd_ (Cwd "_build") ["docker"]
          [ "run", "-it", "alpine"
          , "sh", "-c", "echo HELLO FROM DOCKER"
          ]
    want ["foo"]
```

With Shake 0.18.2 and before, I get:

```
# docker (for foo)
HELLO FROM DOCKER
Build completed in 0.14s
```

But with 0.18.3 and later:

```
# docker (for foo)
```

and then hangs.

gergoerdi:


ndmitchell:
Thanks @gergoerdi for the great bisection and fix work. I've been ill the last few days (and am still ill 😞 ) so will need a few more days to delve into the history that lead to those changes. But it's clear that if something broke Docker, some variation on these changes needs applying!

ndmitchell:
So the reason close_fds defaults to false is that the runtime is linear in the number of open file descriptors, which can be pretty large. The change was introduced in #722 and #727 by @typetetris. Interestingly, it seems that running inside Docker is a particular trigger for this behaviour: https://github.com/ndmitchell/shake/issues/722#issuecomment-564175724

Are both those fields necessary to solve the Docker woes?

gergoerdi:
Sorry I completely forgot about this. I've pushed a new version that only changes `create_group` and still fixes the issue.

gergoerdi:
@ndmitchell can you maybe comment on what that AppVeyor failure means? Looking at the linked log, I don't see what the failure is supposed to be.

gergoerdi:
Something weird is definitely going on with the testsuite:

* At 60935e4b1db609b6d03c7cefde016d74f89bae3b, `stack test` fails with:

```
...
shake> ## BUILD docs
shake> Build completed in 0.01s
shake> 
shake> shake-test: ASSERTION FAILED: Incorrect exception, missing part:
...
shake> TESTS FAILED
shake> Test suite shake-test failed
Test suite failure for package shake-0.19
    shake-test:  exited with: ExitFailure 1
Logs printed to console
$ echo $?
1
```

* With my patch at 1e3738b2a1306afe93d416df2a418b0532350c4c the test suite seems to crash halfway through:

```
shake           > ## BUILD command -j4
...
shake           > # shake_helper (for file)
shake           > # shake_helper (for timeout1)
Progress 2/3: shake
$ echo $?
130
```

gergoerdi:
I think I'm at the limit of my depth of knowledge here on Shake. @ndmitchell can you give at least any initial pointers on why that test could be crashing?

ndmitchell:
I don't think this is necessarily the right answer, but it's better than we have now. So going to merge and iterate from that. I'll update #748 with discussion of what got changed.

PR #753, opened 2020-04-24
Fix negative timings in batch

### LINKED ISSUE #752: Shake report is reporting negative build times for many rules
I've recently turned on Shake profiling for a project, and am a bit surprised to see that it would take a negative amount of time to compile single threaded:

image

Alas, experimental verification concludes that you haven't invented a time machine.

Interestingly, no commands are shown to take a negative time, but plenty of rules are:

image

I can't share the project in question, but I was wondering if you had any thoughts on how I could investigate and if there is further debugging information I could provide.

mavnn:
My attempt to fix #752 . I *think* this addresses the problem, but I am not confident enough in the semantics of how the discounts are supposed to work to be sure.

I'd appreciate your input on this, as we'll be unable to use Shake on CI until we can give meaningful timing information.

### DIFF
--- a/src/Development/Shake/Internal/Core/Action.hs
+++ b/src/Development/Shake/Internal/Core/Action.hs
@@ -546,7 +546,7 @@ batch mx pred one many
             count <- liftIO $ atomicModifyIORef todo $ \(count, bs) -> let i = count+1 in ((i, (b,local,fence):bs), i)
             requeue todo (==) count
             (wait, local2) <- actionFenceRequeue fence
-            Action $ modifyRW $ \root -> addDiscount wait $ localMergeMutable root [local2]
+            Action $ modifyRW $ addDiscount wait
     where
         -- When changing by one, only trigger on (==) so we don't have lots of waiting pool entries
         -- When changing by many, trigger on (>=) because we don't hit all edges


mavnn:
Hi @ndmitchell - just wondering if you'd had a chance to have a look at this, or if there's any one I could speak to who might be able to help?

mavnn:
Although this might be totally on the wrong track! Our linux based CI servers appear to report timing data that makes sense, while running locally on Mac OSX I get the buggy negative numbers. I'll leave this here for your consideration though.

ndmitchell:
Sorry - I had COVID, took a while to recover, and now taking a while to plough through the backlog. Expected to get to Shake stuff on Sunday.

mavnn:
> Sorry - I had COVID, took a while to recover, and now taking a while to plough through the backlog. Expected to get to Shake stuff on Sunday.

Ouch! Sorry to hear that. Take your time to recover!

ndmitchell:
I went through it. The semantics are super subtle, and this patch isn't right, but mostly because there were a series of bugs in this code and you'd have needed to change several things simultaneously to get it working. I've added tests and code, and I'll explain all on the original #752.

PR #755, opened 2020-05-24
Add instance IsCmdArgument (NonEmpty String)

buggymcbugfix:
Add this little useful instance which doesn't introduce any new dependencies.

Please let me know if I should add a test case.

Would it make sense to add `IsCmdArgument` instances for `Text` and `ByteString` as well or were they left out on purpose? (It seems that using `-XOverloadedStrings` isn't really supported anyway?)

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -26,7 +26,9 @@ import Control.Monad.IO.Class
 import Control.Exception.Extra
 import Data.Char
 import Data.Either.Extra
+import Data.Foldable (toList)
 import Data.List.Extra
+import Data.List.NonEmpty (NonEmpty)
 import Data.Maybe
 import Data.Data
 import Data.Semigroup
@@ -38,7 +40,7 @@ import System.IO.Extra hiding (withTempFile, withTempDir)
 import System.Process
 import System.Info.Extra
 import System.Time.Extra
-import System.IO.Unsafe(unsafeInterleaveIO)
+import System.IO.Unsafe (unsafeInterleaveIO)
 import qualified Data.ByteString.Char8 as BS
 import qualified Data.ByteString.Lazy.Char8 as LBS
 import qualified Data.ByteString.UTF8 as UTF8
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks for the patch! Will let CI weigh in, but looks good to me. No need for a test.

Overloaded strings and arguments don't work well, since both ends end up being quite polymorphic. The absence of a ByteString instance is because it's not a very good fit - not many people use ByteString for command line arguments. Text would work well, but I deliberately avoid the dependency on text - the https://hackage.haskell.org/package/shake-plus project might be a better bet for text.

buggymcbugfix:
Ok, it's fair you don't want to depend on text.

Regarding my `Text`/`ByteString` comment—I was suggesting that we might want a `IsCmdArgument` instance for `Text` not for literals, but rather for results of other computations. The point I was trying to make was that shake doesn't seem to support `-XOverloadedStrings` well anyway, hence it might not be too evil to add these instances. My use case is like this:

```hs
foo = do
    Stdout x <- cmd "wibble"
    (xs :: NonEmpty ByteString) <- processBigByteString x
    cmd_ "wuzzle" (map pack $ toList xs)
```

But, hey, I'm happy if I can at least omit the `toList` :)

ndmitchell:
I'm not too fussed about having a ByteString instance for CmdArgument. I guess it would be interpreted as UTF8? While I don't think I'd add one myself, I'd accept such a PR.

buggymcbugfix:
I wouldn't assume any encoding. I'd use [`unpack`](https://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Char8.html#v:unpack).

ndmitchell:
Sure, that's fine too.

bergey:
What encoding does `unpack` implement?  Is it actually Latin-1, or is it a non-standard encoding that covers the same set of characters?  Either way, I'd prefer the type error to shake silently garbling my filenames.

ndmitchell:
See https://hackage.haskell.org/package/bytestring-0.10.10.0/docs/Data-ByteString-Char8.html#v:unpack:

> More specifically these byte strings are taken to be in the subset of Unicode covered by code points 0-255. This covers Unicode Basic Latin, Latin-1 Supplement and C0+C1 Controls.

bergey:
Comparing the [U0080 codepage](http://www.unicode.org/charts/PDF/U0080.pdf) with [Latin-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1) it looks like `unpack` will correctly decode Latin-1 text, as well as accepting some bytes that aren't valid Latin-1.  (I thought Latin-1 assigned meaning to all 256 bytes, but it doesn't.)

buggymcbugfix:
Yes, on second thought I think `Data.ByteString.Char8` is dangerous. I suppose we can assume UTF8, with an appropriate comment that the `StdinBS` option can be used for raw bytestreams. Would this be an acceptable solution?

buggymcbugfix:
I.e. I can't think of any scenario where this might lead to confusing behaviour, but I might be overlooking something.

ndmitchell:
Given there are two possible interpretations, my inclination is not to bake either choice into an instance - people can manually convert to String in whatever locale makes sense for them. Anything else is going to be buggy for someone.

PR #765, opened 2020-06-08
Add neuron

srid:
[neuron](https://neuron.zettel.page/) uses rib (which uses Shake) to manage a static site out of one's Markdown files. Shake is pretty integral to its function!


ndmitchell:
Awesome, thanks! Added to the home page.

Issue #778, opened 2020-10-23
Recursive rules when using AutoDeps on Linux


jacereda:
I've been upgrading a system and noticed that `fsatrace` on `ar` reports a read on the output file right after creating it. 

I started looking into this because shake was complaining with `Rules may not be recursive`. Looking at the code, it seems that the `need` performed in the `AutoDeps` case would need to filter out the output file to avoid the recursive rule.

If you need more info or want me to perform any test just ask.

ndmitchell:
Thanks for the report. The issue will be at https://github.com/ndmitchell/shake/blob/master/src/Development/Shake/Command.hs#L314-L319. I look at FSARead, but it should really be FSARead minus FSAWrite - since a write and a read is usually just a write for build system purposes. Happy to take a patch doing that, or I'll get round to it in the next few days.

ndmitchell:
I've pushed a fix. Let me know if it works for you.

jacereda:
I had hopes I could work on this one this week, but in any case looks like your fix is faster than the naive `read \\ written` I had in mind. I'll try to test it this week and report the results.

jacereda:
Almost there, I had to use `addFSAOptions "rwm"` and include also `[x | FSAMove x _ <- pxs]` in the `written` set.

Sorry for not PRing, that box is a throwaway docker I don't control and I don't have git setup there.

jacereda:
And at this point it looks quite similar to what `fsalint` is doing, so I guess there's some refactoring opportunity here.

ndmitchell:
Anything else to be done here other than a release?

jacereda:
I'll try to test next week, but I think it's enough. I'll just close it and reopen if I find any problem.

PR #780, opened 2020-11-11
Autodeps should consider a rename as a write to the destination path

jacereda:
This is the fix I mentioned in #778 (hopefully, I'm not on linux right now and wrote it from memory).

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -312,8 +312,8 @@ commandExplicitAction oparams = do
             | otherwise = act params
 
         autodeps act = do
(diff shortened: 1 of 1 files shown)


jacereda:
Wow, all builds broken, not bad for 2 lines ;)

jacereda:
Could it be a problem in the CI setups?

ndmitchell:
Yeah, I'd ignore the CI. It's been a bit flakey. I need to invest some time on it , but Travis is shutting down their open source version soon, so I need to migrate it away :(

ndmitchell:
As for the code, looks good - thanks!

PR #789, opened 2021-01-13
addOrOverrideBuiltinRule

pepeiborra:
I want to be able to overwrite previously defined built-in rules, to e.g. customise the behaviour of ghcide.

I will look to add a test once you confirm this contribution will be accepted!

### DIFF
--- a/shake.cabal
+++ b/shake.cabal
@@ -457,6 +457,7 @@ test-suite shake-test
         Test.Batch
         Test.Benchmark
         Test.Builtin
+        Test.BuiltinOverride
         Test.C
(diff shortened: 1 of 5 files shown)


pepeiborra:
Planning to rework this to offer a combinator `overrideBuiltinRule` rather than an option

pepeiborra:
Reworked and tested!

ndmitchell:
Thanks! I renamed the identifier to overrideBuiltinRule (I'm no fan of identifiers that use `Or` in them - it feels really weird to me). If you're happy with that, I'll make a release.

PR #790, opened 2021-01-14
shakeCanRedefineRules: replace overrideBuiltinRule with an option

pepeiborra:
Having an option is nicer in situations with complex systems like ghcide, where Shake is below several layers and not directly exposed. 

This PR also drops `overrideBuiltinRule` to avoid duplication


ndmitchell:
Thanks, looks great! I'll make a release after dinner.

PR #796, opened 2021-03-05
restore depends in newCacheIO

adamse:
We had some problems where rules were not rebuilding, they used caches and this looked like a possible problem.

We're currently testing the fix...

### DIFF
--- a/src/Development/Shake/Internal/Core/Action.hs
+++ b/src/Development/Shake/Internal/Core/Action.hs
@@ -446,6 +446,7 @@ newCacheIO (act :: k -> Action v) = do
             Nothing -> do
                 bar <- newFence
                 pure $ (Map.insert key bar mp,) $ do
+                    Local{localDepends=pre} <- Action getRW
                     Action $ modifyRW $ \s -> s{localDepends = newDepends []}
                     res <- Action $ tryRAW $ fromAction $ act key
                     case res of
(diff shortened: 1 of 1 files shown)


vapourismo:
One underlying issue was also that downgrading was hindered because the Shake Database was corrupted. Building with 0.19.4 also broke previous version 0.19.1.

With this fix, does the Shake database need to be purged/bumped?

adamse:
It looks like this fixed the issue for us.

adamse:
Ping @ndmitchell, could you confirm if I fixed something here or introduced a new problem? :)

ndmitchell:
Thanks for this fix, and sorry it took so long to delve deep into the weeds to figure out exactly what it fixed and why - you were totally right, this was a big screw up. I've pushed a follow up which swaps the order of the arguments to addDepends (to make it consistent whether you hit the cache or have to compute it), and also a test so I don't screw this up in future.

adamse:
Thanks :). I should've mentioned that I found the problem and fix by looking at the diff between the good and the bad version. That made it stand out quite a bit!

adamse:
@ndmitchell could I trouble you for a hackage release with this fix in?

ndmitchell:
Sorry @adamse that this took so incredibly long. Next time, feel free to reach out on my personal email if you've not getting a response. Things have gone crazy personal wise, but I should still manage less than a few months from fix to release...

PR #798, opened 2021-03-13
Optimize away length

pepeiborra:
`firstJustWaitUnordered` contains a complexity "bug": it's runtime is proportional to the number of rule dependencies, whereas we'd much rather have O(rules). 

The profiles that we looked at do not do a good job showing this, but it's unsurprising since the call graph is recursive. Here is the profile that we looked at:

image

And here is another one where I manually inlined `firstJustM` and `maybeM`:

image

This PR only removes an unnecessary call to `length` that can be expensive when the number of elements in the input list is in the thousands.

### DIFF
--- a/src/General/Wait.hs
+++ b/src/General/Wait.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE DeriveFunctor #-}
 
 -- | A bit like 'Fence', but not thread safe and optimised for avoiding taking the fence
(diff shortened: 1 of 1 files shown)


ndmitchell:
Thanks! When you say "bug", I don't see how you could depend on only the number of rules if you walk the graph - that must be approximately O(edges). with the approach Shake takes. (If you have a closed world assumption and order dependencies than O(nodes) is possible - which is approx what Excel does)

pepeiborra:
Agreed, it's a design decision.

PR #806, opened 2021-06-18
Fix documentation for `Stdin`

soenkehahn:
This docstring claimed that `stdin` was by default inherited from the parent process. I think since https://github.com/ndmitchell/shake/commit/db861c3dab925b3cb3e3b6088f2e59461738331f and https://github.com/ndmitchell/shake/commit/102ecfc0c36c5a989db0b1cabad0f4e31a04acf8 that's not true anymore.


ndmitchell:
Thanks!

PR #807, opened 2021-06-20
Fix space leak in modifyAllMem

pepeiborra:
This space leak arises when using reverse deps to avoid work in HLS, but it doesn't seem to impact regular HLS using regular Shake

# Before
I used a cost centre profile to track the leak:
Screenshot 2021-06-20 at 19 20 35

Then I profiled a non-instrumented binary just to verify that the leak was not an artefact of `-prof`. I could still assume that the culprit was still `modifyAllMem` because the type being leaked is `(,)`.
Screenshot 2021-06-20 at 19 21 44

# After 
Screenshot 2021-06-20 at 19 24 04

### DIFF
--- a/src/Development/Shake/Internal/Core/Database.hs
+++ b/src/Development/Shake/Internal/Core/Database.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving, RecordWildCards #-}
 
 module Development.Shake.Internal.Core.Database(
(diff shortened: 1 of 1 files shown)


src/Development/Shake/Internal/Core/Database.hs
```diff
@@ -101,7 +102,9 @@ setMem :: DatabasePoly k v -> Id -> k -> v -> Locked ()
 setMem Database{..} i k v = liftIO $ Ids.insert status i (k,v)
 
 modifyAllMem :: DatabasePoly k v -> (v -> v) -> Locked ()
-modifyAllMem Database{..} f = liftIO $ Ids.forMutate status $ second f
+modifyAllMem Database{..} f = liftIO $ Ids.forMutate status $ \(k,v) ->
```
hasufell, line 105: I wonder, does `status $ \(k, v) -> (k, f v)` also leak?

pepeiborra: It probably does, because tuples are lazy.

hasufell: Well, maybe `second f` blocks inlining, which often is another reason of space leaks.

ndmitchell:
Awesome find! Thanks so much

PR #810, opened 2021-09-04
Add (failing) test for not hashing when we shouldn't

duog:
Hi,

I believe there is a bug in the logic for deciding whether to hash, and I believe these changes demonstrate it.

I've added a `shakeTraceHashFile :: FilePath -> FileHash -> IO ()` to `ShakeOptions` and I call this from inside the `unsafeInterleaveIO` that hashes the file. The digest test maintains an `IORef (HashMap FilePath FileHash)` and asserts it empty after build which should not hash anything(according to the comment).

This is a rough patch, but lmk what you think.

### DIFF
--- a/src/Test/Digest.hs
+++ b/src/Test/Digest.hs
@@ -4,9 +4,14 @@ module Test.Digest(main) where
 import Control.Monad
 import Development.Shake
 import Test.Type
+import Data.HashMap.Strict (HashMap)
+import qualified Data.HashMap.Strict as HashMap
+import Data.IORef
 
-
-main = testBuild test $ do
+main :: IO() -> IO ()
+main sleep = do
+  ref <- newIORef HashMap.empty
+  flip (testBuildTrackHash ref (test ref)) sleep $ do
     want ["Out.txt","Out2.txt"]
 
     "Out.txt" %> \out -> do
@@ -32,20 +37,25 @@ main = testBuild test $ do
         forM_ outs $ \out -> writeFile' out "rewrite"
 
 
-test build = do
+test :: IORef (HashMap FilePath FileHash) -> ([String] -> IO ()) -> IO ()
+test ref build = do
     let outs = ["Out.txt","Out1.txt","Out2.txt"]
     let writeOut x = forM_ outs $ \out -> writeFile out x
     let writeIn = writeFile "In.txt"
     let assertOut x = forM_ outs $ \out -> assertContents out x
+    let clearTrackedHashes = atomicModifyIORef' ref $ \_ -> (HashMap.empty, ())
+    let getTrackedHashes  = atomicModifyIORef' ref $ \x -> (x, x)
+    let assertEmptyHashes = assertNoHashes ref
 
     writeOut ""
     writeIn "X"
     build ["--sleep","--digest-and"]
     assertOut "X"
 
-    -- should not involve a hash calculation (sadly no way to test that)
+    clearTrackedHashes
     build ["--sleep","--digest-and"]
     assertOut "X"
+    assertEmptyHashes
 
     writeIn "X"
     build ["--sleep","--digest-and"]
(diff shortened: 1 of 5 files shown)


src/Test/Digest.hs
```diff
@@ -32,20 +37,25 @@ main = testBuild test $ do
...
     writeOut ""
     writeIn "X"
     build ["--sleep","--digest-and"]
     assertOut "X"
 
-    -- should not involve a hash calculation (sadly no way to test that)
+    clearTrackedHashes
     build ["--sleep","--digest-and"]
     assertOut "X"
+    assertEmptyHashes
```
ndmitchell, line 58: So this is the failing step? The comment says there should be no hash, but you are saying this does perform a hash?

ndmitchell:
Thanks for spotting this! I think 458ecd5101cb4fbe3d3d1506a748283a6616adea fixes this. The issue was that anything in the RunResult is fully forced before being stored in the Shake database (which makes sense), but it noticed that the old and new were 100% equal, that nothing had changed, but then passed over the new value (which had the not-yet-forced thunks for hashes) rather than the old. I've fixed that and added a comment.

I'm not a massive fan of threading through a hook for hashing files, as that's pretty disruptive for a test. But at the same time, untested behaviour can easily regress...

src/Test/Digest.hs
duog: Yes that's right, the test fails here for me.

duog:
> Thanks for spotting this! I think [458ecd5](https://github.com/ndmitchell/shake/commit/458ecd5101cb4fbe3d3d1506a748283a6616adea) fixes this. The issue was that anything in the RunResult is fully forced before being stored in the Shake database (which makes sense), but it noticed that the old and new were 100% equal, that nothing had changed, but then passed over the new value (which had the not-yet-forced thunks for hashes) rather than the old. I've fixed that and added a comment.
> 
> I'm not a massive fan of threading through a hook for hashing files, as that's pretty disruptive for a test. But at the same time, untested behaviour can easily regress...

I agree threading the hook through is gross. I'm sure your fix is good, will verify in a bit, but I'm dubious on the `EqualCost` stuff. For example, https://github.com/ndmitchell/shake/blob/458ecd5101cb4fbe3d3d1506a748283a6616adea/src/Development/Shake/Internal/Rules/File.hs#L162-L163 surely these lines should return `EqualExpensive`.

ndmitchell:
EqualExpensive means that the values are equal, but that it was more expensive than desired to compute that, so you should store a new entry in the database to make future equality cheaper. In both those cases it was expensive to compute, but not any more expensive than we would have expected, and storing a new entry in the database won't make future comparisons cheaper - so they are as cheap as they can be, even if that is still quite expensive.

duog:
> EqualExpensive means that the values are equal, but that it was more expensive than desired to compute that, so you should store a new entry in the database to make future equality cheaper. In both those cases it was expensive to compute, but not any more expensive than we would have expected, and storing a new entry in the database won't make future comparisons cheaper - so they are as cheap as they can be, even if that is still quite expensive.

:+1: that makes a lot of sense, thanks!

duog:
Closing, you fixed the bug (thanks!) and this test is too gross to be merged.

ndmitchell:
Given you found this in reality, would a release including this change be useful?

duog:
I found it hacking on hadrian. Yes, a release would be wonderful :) Thank
you

ndmitchell:
Released as 0.19.6

duog:
Hi, Just following up with some numbers. 

For me, a no-op build of `stage1:exe:ghc-bin` in hadrian speeds up ~ 50%, "+RTS -s" output attached.

I had a thought, perhaps rather than threading a hook through `ShakeOptions` we could report which files were actually hashed (by acting inside the unsafeInterleaveIO) in the profiling report?

[new.stat.txt](https://github.com/ndmitchell/shake/files/7128246/new.stat.txt)
[old.stat.txt](https://github.com/ndmitchell/shake/files/7128247/old.stat.txt)

ndmitchell:
Thanks for the numbers - that's great to see!

Recording which files were hashed in the profile report still requires threading that information somewhere, which is the ugly bit :(

PR #812, opened 2021-09-22
Introduce CommandIOException

mpickering:
This introduces a structured exception for the result of running an
external process. By using a structured type, the build system can catch
these exceptions and control how they are displayed to the user rather
than relying on the default shake formatting.


src/Development/Shake/Command.hs
```diff
@@ -417,15 +417,31 @@ commandExplicitIO params = removeOptionShell params $ \params -> removeOptionFSA
...
-            (if optRealCommand /= optUserCommand then "Original command line: " ++ optUserCommand ++ "\n" else "") ++
-            cwd ++
-            "Exit code: " ++ show (case exit of ExitFailure i -> i; _ -> 0) ++ "\n" ++
-            if null captured then "Stderr not captured because WithStderr False was used\n"
-            else if null exceptionBuffer then intercalate " and " captured ++ " " ++ (if length captured == 1 then "was" else "were") ++ " empty"
-            else intercalate " and " captured ++ ":\n" ++ unlines (dropWhile null $ lines $ concat exceptionBuffer)
+        liftIO $ throwIO $ CommandIOException funcName optRealCommand optUserCommand
+                                              cwd exit captured exceptionBuffer
+
+data CommandIOException =
```
ndmitchell, line 423: If we're going to export this, it ends up being part of the API, so can it have some documentation comments added to it - saying it comes from commands

ndmitchell (commented):
Thanks - shove some docs on and I'll merge!

src/Development/Shake/Command.hs
mpickering: I added the documentation now.

PR #817, opened 2022-01-12
Acknowledge shakeLintIgnore for autodeps

jacereda:
Looks like `AutoDeps` isn't taking `shakeLintIgnore` into account.

### DIFF
--- a/src/Development/Shake/Command.hs
+++ b/src/Development/Shake/Command.hs
@@ -53,6 +53,7 @@ import Development.Shake.Internal.CmdOption
 import Development.Shake.Internal.Core.Action
 import Development.Shake.Internal.Core.Types hiding (Result)
 import Development.Shake.FilePath
+import Development.Shake.Internal.FilePattern
 import Development.Shake.Internal.Options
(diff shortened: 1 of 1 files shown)


jacereda:
I'm not even sure I need this, this was a workaround for the fact that the Visual C linker seems to use the equivalent of `writeFileChanged` for the `IMPLIB:` output file (the .lib when creating a DLL). Looks like `fsatrace` will detect that `.lib` file as a write (when first created or when you change the exported symbols) or as a read (when nothing changed regarding symbols).

I was getting recursive rules that I suspect have to do with that. I'll try to figure out if there's something better on my side (like switching to `&%>` to use the --share) that doesn't require `shakeLintIgnore`.

jacereda:
I guess this could be left as is in favour of a fix for https://github.com/ndmitchell/shake/issues/818

ndmitchell:
Agreed, I think #818 is the right solution here. Broadly speaking I'm not sure linting and auto deps should share settings - they are based on the same underlying mechanism, but are used for quite different purposes.

PR #826, opened 2022-03-23
fix typos

askeblad:
src/Development/Ninja/Type.hs
src/Development/Shake/Database.hs
src/Development/Shake/Internal/Args.hs
src/Development/Shake/Internal/Core/Database.hs
src/Development/Shake/Internal/Rules/Files.hs
src/General/Extra.hs
src/Test/BuiltinOverride.hs


askeblad:
@ndmitchell

src/Test/BuiltinOverride.hs

seems like a newline was added in the commit

ndmitchell:
Thanks! I'm fine with a trailing newline (in fact, I'd prefer it)

PR #835, opened 2023-01-19
doesFileExist_ == True for broken symlinks

1inguini:
I wanted pass the symlink to `produces`, intending to make the target file of the symlink afterwards, but got error: `Files declared by produces not produced:`.

This pull request will let me `produces` broken symlink.


ndmitchell:
Interesting - Shake doesn't really think about or know about symlinks in a meaningful way, and little of the design has considered them. Should symlinks be OK for directories?

I'm inclined to say we merge this, and accept that it's approximate rather than precise, since designing in symlinks from the beginning is a lot of work. But on the fence. Curious if anyone else has opinions.

PR #836, opened 2023-01-21
Add a flag "threaded" for building the executable with the threaded RTS

depressed-pho:
GHC isn't guaranteed to have a threaded RTS. There should be a way to build it with the vanilla one.


ndmitchell:
Thanks!

PR #839, opened 2023-03-23
Add `--watch` option to rebuild when any live file changes

### LINKED ISSUE #371: Consider supporting 'continuous recompilation'
It would be really great to be able to tell shake to keep running, and rebuild any time file changes affect a `need`ed output. I'd like to set up a website asset compilation pipeline that remakes everything upon saving a file, but to do so with shake at the moment I more or less have to duplicate the recompilation rules with another tool that calls shake.

### LINKED ISSUE #570: Build file-watching mechanism into Shake?
I managed to get a `shake watch` working with this:

```haskell
import Control.Concurrent (newChan, readChan)
import Control.Exception (SomeAsyncException(..), fromException, throwIO, try)
import Data.Functor (void)
import Development.Shake
import System.Directory (getCurrentDirectory)
import System.Environment (getArgs, getEnvironment, withArgs)
import System.FSNotify (eventPath, watchTreeChan, withManager)

import qualified Data.Set as Set

main :: IO ()
main =
  -- Get the command line args before shake sees them. This is not ideal because
  -- you must run e.g.
  --
  --   shake watch -j2
  --
  -- as opposed to
  --
  --   shake -j2 watch
  --
  getArgs >>= \case
    "watch":args ->
      -- Strip off the "watch" arg for the real shake script.
      withArgs args $ do
        -- Run the real main under 'try', ignoring synchronous exceptions
        -- because we don't care if the build fails - we want to watch & rebuild
        -- regardless. Re-throw async exceptions.
        try realMain >>= \case
          Left (fromException -> Just (SomeAsyncException ex)) -> throwIO ex
          _ -> pure ()

        -- Get the list of files that shake considers to be alive. This assumes
        -- we've set
        --
        --   shakeLiveFiles = [".shake/live"]
        --
        cwd <- getCurrentDirectory
        files <- Set.fromList . map (cwd </>) . lines <$> readFile ".shake/live"

        -- Start watching the filesystem, and rebuild once any of these files
        -- changes.
        withManager $ \manager -> do
          chan <- newChan
          void (watchTreeChan manager "." ((`elem` files) . eventPath) chan)
          void (readChan chan) -- We block here

        -- Loop. Here my compiled shakefile is itself a build target, so I exec
        -- it rather than loop here, to get the latest & greatest shake
        -- executable.
        env <- getEnvironment
        executeFile "bin/Shakefile" False ("watch":args) (Just env)

    _ ->
      realMain
```

It works okay, but I really don't want to keep writing this code :) 

Should/could something like this live inside shake? `shake --watch`?

chrismwendt:
https://user-images.githubusercontent.com/1387653/227379088-d2bd2aff-1216-448d-b56f-7beadea53ee1.mp4

- Resolves #570 
- Resolves #371


src/Development/Shake/Internal/Args.hs
```diff
@@ -229,43 +232,79 @@ shakeArgsOptionsWith baseOpts userOptions rules = do
...
+                            in case res of
+                                Left err ->
+                                    if Exception `elem` flagsExtra then
+                                        throwIO err
+                                    else do
+                                        putWhenLn Error $ esc Red $ show err
+                                        exitFailure
+                                Right () -> do
+                                    tot <- start
+                                    putWhenLn Info $ esc Green $ "Build completed in " ++ showDuration tot
```
chrismwendt, line 276: Turn off whitespace changes when reviewing this section - it's not as complex as it looks:

CleanShot 2023-03-23 at 16 36 38@2x

PR #844, opened 2023-10-15
Use unsafeUseAsCString where it’s safe to do so

sergv:
Functions in General.Binary mostly read ther contents and don’t care about trailing 0 byte. It’s beneficial to avoid extra copy introduced by regular useAsCString.


ndmitchell:
Thanks, great catch!

PR #849, opened 2024-05-04
Update documentation around non-capturing monadic bind for Action

cdepillabout:
This PR fixes up some of the documentation around the interaction of the non-capturing monadic bind (`>>`) for `Action`, and the `need` `Action`.

From https://neilmitchell.blogspot.com/2019/05/shake-with-applicative-parallelism.html, it appears that since Shake 0.18, `need [foo] >> need [bar]` is intepreted the same way as both `need [foo, bar]`, and `need [foo] *> need [bar]`.  There were a few places left in the Haddocks that weren't yet updated about this, so I've fixed this in this PR.

### DIFF
--- a/src/Development/Shake/Internal/Core/Types.hs
+++ b/src/Development/Shake/Internal/Core/Types.hs
@@ -60,8 +60,14 @@ import Prelude
 --   To raise an exception call 'error', 'fail' or @'liftIO' . 'throwIO'@.
 --
 --   The 'Action' type is both a 'Monad' and 'Applicative'. Anything that is depended upon applicatively
---   will have its dependencies run in parallel. For example @'need' [\"a\"] *> 'need [\"b\"]@ is equivalent
---   to @'need' [\"a\", \"b\"]@.
+--   will have its dependencies run in parallel. For example,
+--   @'Development.Shake.Internal.Rules.File.need' [\"a\"] '*>' 'Development.Shake.Internal.Rules.File.need' [\"b\"]@
+--   is equivalent to @'need' [\"a\", \"b\"]@.
+--
+--   Note that since Shake 0.18, the non-capturing monadic bind ('>>') will also run its dependencies in parallel.
+--   For example,
+--   @'Development.Shake.Internal.Rules.File.need' [\"a\"] '>>' 'Development.Shake.Internal.Rules.File.need' [\"b\"]@
+--   is also equivalent to @'Development.Shake.Internal.Rules.File.need' [\"a\", \"b\"]@.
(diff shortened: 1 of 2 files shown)


src/Development/Shake/Internal/Core/Types.hs
```diff
@@ -60,8 +60,14 @@ import Prelude
 --   To raise an exception call 'error', 'fail' or @'liftIO' . 'throwIO'@.
 --
 --   The 'Action' type is both a 'Monad' and 'Applicative'. Anything that is depended upon applicatively
---   will have its dependencies run in parallel. For example @'need' [\"a\"] *> 'need [\"b\"]@ is equivalent
---   to @'need' [\"a\", \"b\"]@.
+--   will have its dependencies run in parallel. For example,
+--   @'Development.Shake.Internal.Rules.File.need' [\"a\"] '*>' 'Development.Shake.Internal.Rules.File.need' [\"b\"]@
```
cdepillabout, line 64: I needed to fully qualify the path to `need` here, since it appears that Haddock can't figure it out:

https://hackage.haskell.org/package/shake-0.19.8/docs/Development-Shake.html#t:Action

image

See how the `need`s on the last line here aren't actually links, since Haddock doesn't seem to know where it should link to.   Writing these as fully-qualified references should fix this.

src/Development/Shake/Internal/Rules/File.hs
```diff
@@ -371,8 +371,13 @@ fileForward help act = addUserRule $ FileRule help $ fmap ModeForward . act
...
 --
---   Usually @need [foo,bar]@ is preferable to @need [foo] >> need [bar]@ as the former allows greater
---   parallelism, while the latter requires @foo@ to finish building before starting to build @bar@.
+--   Note that the following expressions are all equivalent. @foo@ and @bar@ are built in parallel:
+--
+--   - @need [foo,bar]@
+--   - @need [foo] '*>' need [bar]@
+--   - @need [foo] '>>' need [bar]@
+--
+--   In this situation, @need [foo, bar]@ is preferable, since the parallelism is the most obvious.
```
cdepillabout, line 380: I had also wanted to add a note here saying something like:

> If you do require that two separate `need`s are built sequentially, you can put another (non-`need`) `Action` in between, like the following:
>
> ```haskell
> do
>   need [foo]
>   liftIO (putStr "")
>   need [bar]
> ```

However:

- I wasn't sure what would be the best way to write this (`liftIO putStr` seems kind of hacky)
- This isn't something that _just_ affects 'need', it also affects all the other actions that use `need` internally, like `readFileLines`.

